PR #2, opened 2021-09-01
Added publish to nuget.org action

mjtpena:
- Simple script to publish a nuget package

Some notes:
- need to insert secret to the Security tab
- Manual versioning in place right now (can be automated with SemVer or Git)
- Tests are failing because of socket connections


kjnilsson:
Great, thank you! Ideally at this point I think we should push packages as pre-release (i.e. with a pre or alpha suffix). Also I see you have already pushed a package. Ideally I'd like our packages to be owned by team-rabbitmq on nuget. Do you know how we could arrange this?

kjnilsson:
Ok I'll merge this and make the changes main. Thanks!

PR #8, opened 2021-09-14
Move Endpoint to ClientParameters

CasperTDK:
- [ ] Bump version after? merge


CasperTDK:
I have a docker setup - so I can not use localhost, but have to specific a hostname in my clients

kjnilsson (approved):

Issue #10, opened 2021-09-15
Issues with correlation


CasperTDK:
When I am running a more complex setup with multiple threads or outstanding requests, I am getting correlation issues.

This seems suspicious to me:
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/main/RabbitMQ.Stream.Client/ICommand.cs#L9

Shouldn't CorrelationId always be exposed?

Otherwise I believe https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/main/RabbitMQ.Stream.Client/Client.cs#L253-L261

Will just return the first submitted request response, if there has been a timing issue

CasperTDK:
This has happened in my ClientFactory, where I have bad correlation between my client commands (publish/subscribe) and the QueryMetaData request in the factory.

```
{System.InvalidCastException: Unable to cast object of type 'RabbitMQ.Stream.Client.QueryPublisherResponse' to type 'RabbitMQ.Stream.Client.MetaDataResponse'.
   at RabbitMQ.Stream.Client.Client.QueryMetadata(String[] streams)
   at OCI.DataConsumer.Infrastructure.Installers.StreamClientFactory.GetClient() in C:\projects\omnia.oci\oci.dataconsumer\OCI.DataConsumer\Infrastructure\Installers\StreamClientInstaller.cs:line 52}
```

kjnilsson:
I think these two lines should be using command.CorrelationId instead

https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/main/RabbitMQ.Stream.Client/Client.cs#L280-L281

kjnilsson:
I will push a fix for that

kjnilsson:
Ok pull latests and try

CasperTDK:
OK, let me know when you have bumped the version on nuget :)

CasperTDK:
Wouldn't my MR still be relevant, or am I misunderstanding how correlationIds are used?

kjnilsson:
Because the interface exposes a default CorrelationId value we don't need to implement it for every ICommand, only those that are actually correlated so I don't think your PR fixes anything.

kjnilsson:
are we ok with this issue after the fix?

CasperTDK:
Let me double check tonight - roughly in 3 hours

CasperTDK:
@kjnilsson Your fix did the trick!

PR #11, opened 2021-09-15
Ensured CorrelationId is always exposed on ICommand

CasperTDK:
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/10


CasperTDK:
@kjnilsson  rebased

PR #12, opened 2021-09-15
Bug: ConsumerShouldReceiveDelivery_MultipleChunks fails

CasperTDK:
While testing correlations, I discovered I only get 1 chunk in my deliverHandler. Even though I have multiple chunks in my stream.
I have added a test case to showcase the issue

### DIFF
--- a/Tests/ClientTests.cs
+++ b/Tests/ClientTests.cs
@@ -110,7 +110,7 @@ namespace Tests
                 testOutputHelper.WriteLine($"publish error code ${errors[0]}");
                 testPassed.SetResult(true);
             };
-            
+
             var publisherRef = Guid.NewGuid().ToString();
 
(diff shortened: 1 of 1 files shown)


kjnilsson:
try setting initialCredit=2, the credit flow is at the chunk level so you've only provided enough credit to receive 1 chunk.

CasperTDK:
Ah thanks, can you link to a place that describes the credit flow? I have been unsure about this

kjnilsson:
I don't think we have one apart from [this](https://github.com/rabbitmq/rabbitmq-server/blob/v3.9.x/deps/rabbitmq_stream/docs/PROTOCOL.adoc#credit)

in your test you can also use client.Credit(subId, credit) to give more credit. 

I good approach is to give at least two chunks worth of credit then give 1 more credit after each processed chunk.

CasperTDK:
Thanks again. It is much appreciated

PR #15, opened 2021-10-05
Allocation reductions and CPU/Task optimizations

stebet:
* Removed channels from Client and Connection. This reduced the number of tasks that need be running and gets rid of double-buffering for frames sent/received.
* Added a bounded channel on the Producer to batch `Publish` messages sent.
* Made the Request operations generic so it's now `Request<TIn, TOut>`
* Added a pooled `IValueTaskSource` to reuse `ValueTask` objects for Request operations.
* Added generic optimizations for the `Connection.Write` method for better code generation and to eliminate struct boxing for ICommand objects.
* Added array pooling for `ulong` arrays in message confirmations and for frame handling. Arrays are returned after confirmations have been processed and after frames have been handled.


stebet:
Trying to figure out why that test is flaky.

kjnilsson:
I think there is a fix for that in main

stebet:
> I think there is a fix for that in main

No this is the DeliverOffset test that is failing. I'm taking a closer look at it.

kjnilsson (approved):

Issue #49, opened 2022-01-01
Change NuGet behaviour on github action  


Gsantomaggio:
- Remove NuGet push during the PR. It is not needed and often it raises an exception due to the duplication version. 
- Add NuGet Push when TAG new version.

PR #50, opened 2022-01-03
Split Test and NuGet Push

Gsantomaggio:
nuget.yaml: the nuget push is executed only when there is a TAG
test.yaml: it does not contain the nuget push anymore

Closes https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/49


Gsantomaggio:
@mjtpena What do you think?

PR #56, opened 2022-01-10
Add producer MessagesBufferSize

Gsantomaggio:
Add tests and validation for stream name and MessagesBufferSize  limit


kjnilsson:
Should this be MaxBatchSize given the batch size can be smaller

lukebakken (approved):

PR #60, opened 2022-01-26
Amqp 1.0 Implementation

### LINKED ISSUE #58: AMQP 0.9.1 interoperability
Hi!

We're trying to consume messages from RabbitMQ streams using the stream-plugin and this dotnet stream client. It works fine until someone publishes a message through either an AMQP 0.9.1 channel or by using the management UI.

Once the stream consumer receives the AMQP 0.9.1 message, it seems to freeze, and will no longer trigger the "MessageHandler" of the registered consumer.

Am I missing something? Or is it just not yet implemented?

Steps to reproduce:
1. Hook up a stream-consumer to a stream
2. Publish a message using the stream protocol
3. Make sure the message is received
4. Publish a message to the same stream using the management UI
5. The stream consumer will no longer trigger MessageHandler delegate

Gsantomaggio:
Fixes #58 

AMQP 1.0 Message implementation 
---

Disclamer:
The implementation is not completed yet. 
At the moment it supports:
- Body
- Annotations
- ApplicationProperties
- Properties

that are the most used attributes.

Code:
---
Most of the new code is inside the `AMQP` directory where you can find:
- ` AmqpWireFormatting` ( divided in `AmqpWireFormattingRead` and `AmqpWireFormattingWrite` partial classes)
 The class is responsible to read/write AMQP fields with the format codes
- `FormatCode` with all the FormatCodes

How to test:
 - send messages from another stream client (GO, Java) with the attributes 
 - send amqp 091 messages

Unit tests:
- I generated some use cases with the goclient and put the [dump as resources](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/60/files#diff-8b8ee0617579680e53b9d8a2866163ce11a5f7d1908e43a8723781a48de6bce2R296) so I can test the decode from different sources
- I saw some test code form [amqp-net-lite](https://github.com/Azure/amqpnetlite/blob/master/test/Test.Amqp.Net/AmqpCodecTests.cs#L32) and I did something [similar](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/60/files#diff-8b8ee0617579680e53b9d8a2866163ce11a5f7d1908e43a8723781a48de6bce2R120)
 - there is also a test with [AMQP 091](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/60/files#diff-f8e8a02852eff4dc4c89bb97a257c59c566e83128fc12529e44c913321957535R411)


RabbitMQ.Stream.Client/AMQP/Described.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace RabbitMQ.Stream.Client.AMQP
+{
+    public readonly struct Described
```
MarcialRosales, line 8: I would rename this class to be DataCode and then methods just Write and Read.

RabbitMQ.Stream.Client/AMQP/FormatCode.cs
```diff
@@ -0,0 +1,98 @@
+namespace RabbitMQ.Stream.Client.AMQP
+{
+    public static class FormatCode
```
MarcialRosales, line 3: Would it be possible to actually merge this constants into above class, `DescribedFormatCode`? I mean, rather than having two classes.

RabbitMQ.Stream.Client/AMQP/Map.cs
```diff
@@ -0,0 +1,85 @@
...
+                _ => false
+            };
+        }
+
+        public int Size
+        {
+            get
+            {
+                var size = Described.Size;
+                size += 1; //FormatCode.List32
```
MarcialRosales, line 60: Maybe use constants instead of actual values?

RabbitMQ.Stream.Client/Message.cs
```diff
@@ -1,29 +1,58 @@
...
+            if (Annotations != null)
+            {
+                offset += Annotations.Write(span.Slice(offset));
+            }
+
+            offset += Data.Write(span.Slice(offset));
+            return offset;
         }
 
         public ReadOnlySequence<byte> Serialize()
```
MarcialRosales, line 58: This method does not seem to be referenced. is it needed?

RabbitMQ.Stream.Client/Message.cs
```diff
@@ -36,9 +65,55 @@ public ReadOnlySequence<byte> Serialize()
...
         {
+            //                                                         Bare Message
+            //                                                             |
+            //                                       .---------------------+--------------------.
+            //                                      |                                           |
+            // +--------+-------------+-------------+------------+--------------+--------------+--------
+            // | header | delivery-   | message-    | properties | application- | application- | footer |
+            // |        | annotations | annotations |             | properties  | data         |        |
+            // +--------+-------------+-------------+------------+--------------+--------------+--------+            
+
```
MarcialRosales, line 76: May I suggest to add these comments? I think they are useful:
 //  Altogether a message consists of the following sections:
            // • Zero or one header.
            // • Zero or one delivery-annotations.
            // • Zero or one message-annotations.
            // • Zero or one properties.
            // • Zero or one application-properties.
            // • The body consists of either: one or more data sections, one or more amqp-sequence sections,
            // or a single amqp-value section.
            // • Zero or one footer.

RabbitMQ.Stream.Client/AMQP/Properties.cs
```diff
@@ -0,0 +1,152 @@
+using System;
+using System.Buffers;
+using System.Runtime.InteropServices;
+
+namespace RabbitMQ.Stream.Client.AMQP
+{
+    public class Properties : IWritable
```
MarcialRosales, line 7: I would unit test this class. Same comment I did earlier for AmqpWireFormatterWriter applies here.

RabbitMQ.Stream.Client/Message.cs
Gsantomaggio: do you mean `Serialize()` ?

Atm is not used yet. At some point we will use it

MarcialRosales (commented):
The only comment to highlight because the PR is good to go as it is is maybe to unit tests the key classes like various parsers, and utility parsers and the publish and deliver commands. I believe by having unit tested these classes we can make the other tests cases which are more collaboration tests simpler with less data assertions because we know that the serialization works in all scenarios. 
Other than that, it looks good to me.

Issue #83, opened 2022-03-12
OffsetTypeOffset: Start consuming from previous messages than stored offset


simone-fariselli:
Using OffsetTypeOffset, once you stored an offset the consumer has to start consuming from the stored offset for the reference. It starts from a lower offset value.

PR #84, opened 2022-03-12
Consumer start from stored offset

simone-fariselli:
- Handling start from stored offset
- Implemented test for feature: ShouldConsumeFromStoredOffset in ConsumerSystemTests.cs


Gsantomaggio:
Grazie @simone-fariselli 
This is the fix for https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/83

lukebakken (approved):

PR #86, opened 2022-03-16
removed public access modifiers

sirh3e:
Removed public access modifiers to match other interfaces like ICommand, ILogEventSource, INamedEntity


sirh3e:
Have look at #89

PR #88, opened 2022-03-16
use of double deconstruction instead of single twice

sirh3e:
Use of double deconstruction instead of single twice


sirh3e:
Have look at #89

Issue #64, opened 2022-02-15
Handle MetaData update 


Gsantomaggio:
Raise an event when the MetaData are updated. 
Meta Data update is raised when:
 - a stream is deleted 
 - A leader is removed 

The producer and consumer should have an event to handle it.

0UserName:
> Raise an event when the MetaData are updated.

Will the event be a replacement for `ClientParameters.MetadataHandler`?

Gsantomaggio:
> Will the event be a replacement for ClientParameters.MetadataHandler?

No, I'd like to expose the event on `ProducerConfig` and `ConsumerConfig`.
ATM if MetadataHandler is raised the producers and consumers don't react properly in this event.
Need to investigate btw I am not sure about the right strategy.

0UserName:
May I ask what is 'react properly'? Let's say the metadata was updated due to the loss of the leader or the deletion of the stream. How should the producer and consumer handle this event?

acogoluegnes:
The client should use the `metadata` command to find out about the state of the stream (does not exist anymore, location of leader and replicas) and decide what's to do depending on the response (connect to the node of the leader for a publisher, connect to a node where there's a replica, or fail if the stream has been deleted).

Gsantomaggio:
.NET client does not implement the Autoreconnet (yet). 
It is useful to notify the user that the stream topology changed or the stream is deleted. 

When you delete a stream, In the Goclient  there is something like:
```golang
producer, err := env.NewProducer(streamName, nil)
cbClose := producer.NotifyClose()
go func(ch stream.ChannelClose) {
      event := <-ch
      fmt.Printf("Producer closed. Stream: %s, Reason: %s", event.StreamName, event.Reason)
}(cbClose)
```
and you receive the event:
```
Producer closed. Event Name: , Stream: meta_data_test, Reason: Metadata update
```

The Java client is smarter and can auto-reconnect automatically. 


>How should the producer and consumer handle this event?

ATM raise an event maybe using:
```csharp
  ConnectionClosedHandler = s =>
    {
        Console.WriteLine($"ConnectionClosedHandler: {s}");
        return Task.CompletedTask;
     },
```

But It needs more investigation

Gsantomaggio:
Ok, the `Producer/Consumer` classes should only expose an event. 
The new [Reliable classes](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/92)  will react to this event

PR #93, opened 2022-03-28
Handle metadata

Gsantomaggio:
closes https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/64


lukebakken (changes requested):
APIApproval needs to be updated and it looks like there may be one flaky test.

Gsantomaggio:
@lukebakken file updated! I always forget it :) !

PR #99, opened 2022-04-02
QueryOffset and QuerySequence

simone-fariselli:
Add QueryOffsetWithDefaultValue
Add tests
Change in QueryOffset and QuerySequence: on reference null or empty throws as discussed with @Gsantomaggio 
Refactor


RabbitMQ.Stream.Client/StreamSystem.cs
```diff
@@ -111,6 +111,18 @@ public async Task<bool> StreamExists(string stream)
             return response.StreamInfos is { Count: >= 1 } &&
                    response.StreamInfos[stream].ResponseCode == ResponseCode.Ok;
         }
+        private static void MaybeThrowQueryException(string reference, string stream)
+        {
+            if (string.IsNullOrEmpty(reference) || string.IsNullOrEmpty(stream))
+            {
+                throw new QueryException("Stream name can't be empty");
```
Gsantomaggio, line 118: I would change in:
```csharp
QueryException("Stream name reference  and can't be empty");
````

File: RabbitMQ.Stream.Client/StreamSystem.cs [hunk 2]
```diff
@@ -121,17 +133,32 @@ public async Task<bool> StreamExists(string stream)
...
+
+        /// <summary>
+        /// QueryOffsetWithDefaultValue retrieves the last consumer offset stored
+        /// given a consumer name and stream name 
+        /// or 0 as default value if thre is no stored value
+        /// </summary>
+        /// <param name="reference">Consumer name</param>
+        /// <param name="stream">Stream name</param>
+        /// <returns></returns>
+        public async Task<ulong> QueryOffsetWithDefaultValue(string reference, string stream)
```
Gsantomaggio, line 152: I would add an overload for `QueryOffset(string reference, string stream, ulong defaultValue)`

Gsantomaggio (changes requested):

RabbitMQ.Stream.Client/StreamSystem.cs
simone-fariselli: Ok, 
I will set the message "Stream name and reference can't be empty or null".

simone-fariselli:
Removed QueryOffsetWithDefaultValue as discussed with @Gsantomaggio

Gsantomaggio (approved):

Gsantomaggio:
thank you @simone-fariselli

PR #100, opened 2022-04-04
Change the amqp 1.0 test

Gsantomaggio:
MessageId should be an uint64 in this case. Add validation for it.
@wolf4ood you can use the new `Tests/Resources/message_random_application_properties_properties_900` to validate the RUST amqp1.0 parser. Thank you


wolf4ood:
LGTM, Rust test are now passing with payload `message_random_application_properties_properties_900` :)

Gsantomaggio:
Will merge per @wolf4ood approves

PR #108, opened 2022-04-21
Documentation

simone-fariselli:
added a bit of documention on offset types

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ var consumer = await system.CreateConsumer(
 });
 ```
 ### Offset Types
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -296,6 +297,45 @@ var consumer = await system.CreateConsumer(
         }
 });
 ```
+### Offset Types
+There are five types of Offset and can be set by the property  `OffsetSpec` of the class `ConsumerConfig` that must be passed to the Consumer constructor, in the example we use `OffsetTypeFirst`:
```
sungam3r, line 301:
```suggestion
There are five types of Offset and they can be set by the `ConsumerConfig.OffsetSpec` property that must be passed to the Consumer constructor, in the example we use `OffsetTypeFirst`:
```

Gsantomaggio:
Thank you @simone-fariselli @sungam3r

PR #121, opened 2022-05-17
Handle infinite looping when encountering a misconfigured load balancer

ricsiLT:
Proposal to handle infinite looping when connecting to a load balancer where nodes are misconfigured, causing client to hang indefinitely.

Logic can be replaced by time-based loop, separate exception could be used if necessary (maybe even exposing nodes that were tried for easier diagnostics).

### DIFF
--- a/RabbitMQ.Stream.Client/RoutingClient.cs
+++ b/RabbitMQ.Stream.Client/RoutingClient.cs
@@ -80,15 +80,20 @@ namespace RabbitMQ.Stream.Client
             var advertisedHost = GetPropertyValue(client.ConnectionProperties, "advertised_host");
             var advertisedPort = GetPropertyValue(client.ConnectionProperties, "advertised_port");
 
-            while (broker.Host != advertisedHost ||
-                   broker.Port != uint.Parse(advertisedPort))
+            var attempts = 0;
+            while (broker.Host != advertisedHost || broker.Port != uint.Parse(advertisedPort))
             {
-                await client.Close("advertised_host or advertised_port doesn't mach");
+                attempts++;
+                await client.Close("advertised_host or advertised_port doesn't match");
 
                 client = routing.CreateClient(clientParameters with { Endpoint = endPoint });
 
                 advertisedHost = GetPropertyValue(client.ConnectionProperties, "advertised_host");
                 advertisedPort = GetPropertyValue(client.ConnectionProperties, "advertised_port");
+                if (attempts > 100)
+                {
+                    throw new RoutingClientException($"Could not find broker ({broker.Host}:{broker.Port}) after 100 attempts");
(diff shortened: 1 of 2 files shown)


RabbitMQ.Stream.Client/RoutingClient.cs
```diff
@@ -80,15 +80,20 @@ string GetPropertyValue(IDictionary<string, string> connectionProperties, string
             var advertisedHost = GetPropertyValue(client.ConnectionProperties, "advertised_host");
             var advertisedPort = GetPropertyValue(client.ConnectionProperties, "advertised_port");
 
-            while (broker.Host != advertisedHost ||
-                   broker.Port != uint.Parse(advertisedPort))
+            var attempts = 0;
```
Gsantomaggio, line 83: Instead of a fixed value, we can use the `StreamInfo`.

So add a parameter to `LookupConnection(ClientParameters clientParameters, Broker broker)` something like: `LookupConnection(ClientParameters clientParameters, Broker broker, int attempts)` 


When used:
```csharp
public static async Task<IClient> LookupLeaderConnection(ClientParameters clientParameters,
            StreamInfo metaDataInfo)
        {
     
       var attempts =  1 + // this is the leader
            metaDataInfo.Replicas.Count // number of replicas 
    
        attempts = attempts *2; // just to be sure about loadbalancer redirect

       return await LookupConnection(clientParameters, metaDataInfo.Leader, attempts);
        }
```

Gsantomaggio (changes requested):

Gsantomaggio:
Thank you for your PR. [I left a comment ](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/121#discussion_r874708045)

ricsiLT:
Submitted changes. Tried running unit tests with your approach (`(1 + replicas.Count) * 2`) but it would spuriously fail. Tell me if `Pow()` is too much :D

Gsantomaggio:
Can you please execute `dotnet format` and then commit again? Thank you

ricsiLT:
Sorry for multiple commits, can only do this PR via github web UI.

Gsantomaggio:
@ricsiLT don't worry :)! We will Squash the commits!

Gsantomaggio:
In order to format you need to use the last version:
```
$ dotnet --version
6.0.300
```

they changed something, so previous versions won't work.

if you can't update `dotnet`, please give me the grant I will format the code.

ricsiLT:
So you'll merge this whenever? Nothing else required from me?

Gsantomaggio:
>So you'll merge this whenever? 

I formatted the code. I will merge! Thank you

Gsantomaggio (approved):

PR #135, opened 2022-06-13
Fix a bug with Heartbeat values >59s

ricsiLT:
Noticed that our heartbeat was 0 when we set it to 60. This fixes that.
Not sure if it's totally okay since timespans can be negative and casting negatives to `uint` will produce big numbers. Anyhow, maybe that's an edge case that shouldn't be covered :D


ricsiLT:
https://dotnetfiddle.net/yL6J8M if you want to confirm :)

Gsantomaggio:
@ricsiLT oh! thank you!

PR #139, opened 2022-06-21
Improved client timeout handling

ricsiLT:
Make timeout configurable
Add some descriptions to enum cases

### DIFF
--- a/RabbitMQ.Stream.Client/Reliable/ReliableProducer.cs
+++ b/RabbitMQ.Stream.Client/Reliable/ReliableProducer.cs
@@ -11,13 +11,28 @@ namespace RabbitMQ.Stream.Client.Reliable;
 
 public record ReliableProducerConfig
 {
+    private readonly TimeSpan _timeoutMessageAfter = TimeSpan.FromSeconds(3);
+
     public StreamSystem StreamSystem { get; set; }
     public string Stream { get; set; }
     public string Reference { get; set; }
     public Func<MessagesConfirmation, Task> ConfirmationHandler { get; init; }
     public string ClientProvidedName { get; set; } = "dotnet-stream-rproducer";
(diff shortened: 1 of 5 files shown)


Gsantomaggio:
Can you please update the documentation? https://github.com/rabbitmq/rabbitmq-stream-dotnet-client#invalidate-messages. 

Maybe explain the ClientSide exception vs server-side exception (`ClientTimeoutError`)

RabbitMQ.Stream.Client/Reliable/ReliableProducer.cs
```diff
@@ -17,6 +17,7 @@ public record ReliableProducerConfig
     public Func<MessagesConfirmation, Task> ConfirmationHandler { get; init; }
     public string ClientProvidedName { get; set; } = "dotnet-stream-rproducer";
     public IReconnectStrategy ReconnectStrategy { get; set; } = new BackOffReconnectStrategy();
+    public TimeSpan TimeoutMessageAfter { get; init; } = TimeSpan.FromSeconds(3);
```
Gsantomaggio, line 20: Should we add a sort of validation to avoid to insert wrong values like `0` or `Hours` ?

ricsiLT: No idea - 0 doesn't make sense for sure, so we can definitely check for that. For other values - other than just wasting space in cache, what's the danger on waiting for days on end? :D

Gsantomaggio: > what's the danger on waiting for days on end? :

The unconfirmed list is in memory. Just a memory problem. I'd say!

Gsantomaggio:
Thank you @ricsiLT

PR #142, opened 2022-06-27
Batch send

### LINKED ISSUE #136: Request for awaitable batch publish
Hi,

We are implementing a publisher extension like that right now but maybe it would be useful for more users of the library.

The only firm requirements we have is 

* every message has to have a publishing id (sub-batching isn't an option for us :( )
* we consider failing to deliver any message as a failure/exception

In our implementation, the usage looks more or less like this:

```csharp
var cts = new CancellationTokenSource();
var token = cts.Token;

// reference for producer would be either random or null, since we envision such producer to be disposable.
// Also, it's more simple to think about such producer as single-use
//     * then you don't have to worry about calling `PublishAllMessages` multiple times
//     * and thinking about publishing ids
await using var x = await system.CreateBatchProducer(streamName); 

// publishing should
//     * throw if at least single message failed (not sure how retrying would work - maybe it's possible?)
//     * cancellable
await x.PublishAllMessages(messagesToPublish, token);
```

But since this would be library code, we don't want to push any specific impl, just showcasing how we aim to use it now.

Is it something you'd be looking to add?

Gsantomaggio:
Closes #136

Using this gist: https://gist.github.com/Gsantomaggio/cd5b0734225d3fab9ed013d634eaf005 to test the performances 

cc @ricsiLT 

Test results `Producer` and `Reliable Producer` with and without `Batch Send` :
```
*****Standard Producer Batch Send***** send time: 00:00:24.3330340, messages sent: 20000000

*****Standard Producer Send***** send time: 00:00:24.6326600, messages sent: 20000000

*****Reliable Producer Batch Send***** time: 00:00:27.2689740, messages sent: 20000000

*****Reliable Producer Send No Batch***** send time: 00:00:31.4620940, messages sent: 20000000
```

These values are excepted the good news is `Reliable Producer Batch Send` time is not so far from standard producer with `batch send`

### DIFF
--- a/Examples/Performances/BatchVsBatchSend.cs
+++ b/Examples/Performances/BatchVsBatchSend.cs
@@ -1,0 +1,245 @@
+using RabbitMQ.Stream.Client;
+using RabbitMQ.Stream.Client.Reliable;
+
+namespace Performances;
+
+public class BatchVsBatchSend
+{
+    private const int TotalMessages = 20_000_000;
+    private const int MessageSize = 100;
+    private const int AggregateBatchSize = 300;
+    private const int ModPrintMessages = 10_000_000;
+
+    public async Task Start()
+    {
+        Console.WriteLine("Stream Client Performance Test");
+        Console.WriteLine("==============================");
+        Console.WriteLine("Client will test Batch vs Batch Send");
+        Console.WriteLine("Total Messages: {0}", TotalMessages);
+        Console.WriteLine("Message Size: {0}", MessageSize);
+        Console.WriteLine("Aggregate Batch Size: {0}", AggregateBatchSize);
+        Console.WriteLine("Print Messages each: {0} messages", ModPrintMessages);
+
+
+        var config = new StreamSystemConfig() {Heartbeat = TimeSpan.Zero};
+        var system = await StreamSystem.Create(config);
+        await BatchSend(system, await RecreateStream(system, "StandardBatchSend"));
+        await StandardProducerSend(await RecreateStream(system, "StandardProducerSendNoBatch"), system);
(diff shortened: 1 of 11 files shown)


Gsantomaggio:
Using this gist: https://gist.github.com/Gsantomaggio/cd5b0734225d3fab9ed013d634eaf005 to test the performances

Gsantomaggio:
@ricsiLT when you have time, can you please test it?

ricsiLT:
Just built it, will deploy to syst shortly ^^

Gsantomaggio:
Moved the Performances to https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/tree/batch_send/Examples/Performances

ricsiLT:
Seems to be working fine. Would only be able to test with huge loads next week, but small publishes look ok.

ricsiLT:
Hmmm, in the end I did get some errors:

```
System.InvalidOperationException: Total size of messages in batch is too big. Max allowed is 1048576
   at RabbitMQ.Stream.Client.Producer.PreValidateBatch(List`1 messages)
   at RabbitMQ.Stream.Client.Reliable.ReliableProducer.BatchSend(List`1 messages)
```

Should this be handled on my end? Or should function split it into multiple batches? How can I change that limit?

Gsantomaggio:
> Should this be handled on my end?

Yes

 > Or should function split it into multiple batches? 
 
 This is what the standard `Send` does internally so It doesn't make sense to add a split here else we have another `Send` 

> How can I change that limit?

It is possible to request a higher value but atm the API is missing. I will add

ricsiLT:
Can confirm that sending small batches works, and validation for batches that are too big seems to work as well :D merge it in!

Issue #150, opened 2022-07-19
ReliableProducer not reconnecting after leader goes down


ricsiLT:
Hi, we observe that ReliableProducer does attempt to reconnect cluster after leader goes down.

Our setup:

* 3 node cluster (Node_A, Node_B, Node_C, let's say that Node_A is the leader)
* no load balancer

When we restarted the leader, we observed the following in the logs:

```
INFO: Producer reference: X, stream: Y  disconnected, check if reconnection needed in 200 ms.
ERROR: Error during producer initialization: System.Net.Sockets.SocketException (10061): No connection could be made because the target machine actively refused it.
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
   at System.Threading.Tasks.ValueTask.ValueTaskSourceAsTask.<>c.<.cctor>b__4_0(Object state)
--- End of stack trace from previous location ---
   at RabbitMQ.Stream.Client.Connection.Create(EndPoint endpoint, Func`2 commandCallback, Func`2 closedCallBack, SslOption sslOption)
   at RabbitMQ.Stream.Client.Client.Create(ClientParameters parameters)
   at RabbitMQ.Stream.Client.StreamSystem.MayBeReconnectLocator()
   at RabbitMQ.Stream.Client.StreamSystem.CreateProducer(ProducerConfig producerConfig)
   at RabbitMQ.Stream.Client.Reliable.ReliableProducer.GetNewReliable(Boolean boot)
 ---> System.Net.Sockets.SocketException Data:
    Host : Node_A
    Port : 5551
```

After that, no reconnection was attempted by reliable producer.

Gsantomaggio:
Thank you for reporting. 
That part _has_  to be improved.

Gsantomaggio:
Ok I can reproduce the issue with three nodes cluster and this configuration:
```csharp
 var config = new StreamSystemConfig()
        {
            UserName = "test",
            Password = "test",
            Endpoints = new EndPoint[]
            {
                new DnsEndPoint("node0", 5552),
                new DnsEndPoint("node1", 5552),
                new DnsEndPoint("node2", 5552),
            }
        };
```

Then stop the leader node

PR #151, opened 2022-07-19
Change reconnection logic in ReliableBase

ricsiLT:
In response to #150

### DIFF
--- a/RabbitMQ.Stream.Client/PublicAPI.Unshipped.txt
+++ b/RabbitMQ.Stream.Client/PublicAPI.Unshipped.txt
@@ -639,11 +639,11 @@ RabbitMQ.Stream.Client.Reliable.MessagesConfirmation.PublishingId.get -> ulong
 RabbitMQ.Stream.Client.Reliable.MessagesConfirmation.Status.get -> RabbitMQ.Stream.Client.Reliable.ConfirmationStatus
 RabbitMQ.Stream.Client.Reliable.ReliableBase
 RabbitMQ.Stream.Client.Reliable.ReliableBase.Init() -> System.Threading.Tasks.Task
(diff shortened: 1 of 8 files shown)


lukebakken:
Thank you for your continued contributions. I just returned to work after some PTO, and @Gsantomaggio is out for a bit. We'll get to this when we can!

lukebakken:
@ricsiLT thanks for your patience, I'm reviewing this today.

lukebakken (changes requested):
There are 4 test failures that do not happen on `main`. I'm investigating.

ricsiLT:
Yeah, test failures puzzled me too since I didn't understand why they wouldn't fail on `main` in the first place - chatted about it with @Gsantomaggio on Slack but he was busy with something else so I didn't get a clear answer about those tests :)

lukebakken:
@Gsantomaggio when you get back, let's discuss the remaining test failure. When running the `ConsumeAfterKillConnectionShouldContinueToWork` test we can see that many connections are killed at the end of the test. I'm wondering if the new reconnection logic creates many new connections?

[rabbitmq-stream-dotnet-client-151-failure-log.txt](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/files/9293545/rabbitmq-stream-dotnet-client-151-failure-log.txt)

I'm going to look into how to enable logging while running tests. Not sure how to do that now.

Gsantomaggio:
@ricsiLT Thanks for the PR!
I didn't check the code, but there must be a bug somewhere. 
a lot of connections are recreated 
Screen Shot 2022-08-12 at 09 08 24


This is why the assert `Assert.Equal(1, connectionsKilled);` fails.

Gsantomaggio:
@ricsiLT Thank you again for your contribution. I close this PR in favour of #163 !

Issue #157, opened 2022-08-21
 GetAnySize doesn't handle double value


Gsantomaggio:
the function `public static int GetAnySize(object value)` does't handle the double value

Gsantomaggio:
How to reproduce:
```csharp
double dv = 6665555.34566;
 var app = new ApplicationProperties()
            {
                ["double_value"] =dv
            };

            var m = new Message(Encoding.Default.GetBytes(strJson))
            {
                ApplicationProperties = app
            };

            await _producer.Send(m);
```

It raises an exception

PR #158, opened 2022-08-22
Handle more type in the AMQP 1.0 GetSize

Gsantomaggio:
Fixes: #157 

Add Size Value for:
- bool, 
- short, 
- float 
- double 

on the AMQP 1.0 codec.

Add more tests for the Size Validation


Tests/Amqp10Tests.cs
```diff
@@ -374,6 +371,56 @@ public void ValidateMessagesFromGo()
...
+            size += AmqpWireFormatting.GetAnySize("long_value");
+            size += AmqpWireFormatting.GetAnySize(1L);
+            size += AmqpWireFormatting.GetAnySize("ulong_value");
+            size += AmqpWireFormatting.GetAnySize(1UL);
+            size += AmqpWireFormatting.GetAnySize("float_value");
+            size += AmqpWireFormatting.GetAnySize(1.0f);
+            size += AmqpWireFormatting.GetAnySize("date_value");
+            size += AmqpWireFormatting.GetAnySize(dt);
+
+            Assert.Equal(size, m.ApplicationProperties.Size);
```
Zerpet, line 421: I'm not very enthusiastic about this test case. The issue with aggregating all the sizes and asserting on the aggregate, is that you don't know if the individual size for each type is correct.

For example, I can "break" this test by swapping the value of `GetFloatSize()` and `GetDoubleSize()`. The test is still green, but the value returned by those two functions is be incorrect.

I think we must assert on each individual call to `GetAnySize()` to guarantee that we are returning the right thing.

Gsantomaggio: You are right [Done here](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/158/commits/32439232382b49f0127f30a4b42dc741699e6375)

Zerpet (approved):

Issue #154, opened 2022-07-29
Error during close consumer during the delivery


Gsantomaggio:
Bug found by  @ricsiLT

When `consumer.close()` raises an exception when there are still some deliver frames on the network. 

Also, the client receives `throw new ArgumentException($"Unknown or unexpected tag: {tag}", nameof(tag));` for tag 9.

To reproduce the issue, it is enough to close the client when it is consuming a large number of messages for ex `5_000_000`:
```
var c2 = await ReliableConsumer.CreateReliableConsumer(new ReliableConsumerConfig()
        {
            StreamSystem = system,
            Stream = streamName,
            Reference = applicationName,
            ClientProvidedName = "second",
            OffsetSpec = new OffsetTypeFirst(),
            MessageHandler = async (c, context, arg3) =>
            {
                Thread.Sleep(1);

           
            }
        });

            Console.WriteLine("closing");
            Thread.Sleep(10);
             await c1.Close();
            
```

ricsiLT:
Logs on client side that indicate this issue:

```csharp
Error removing the consumer id: 0 from the server. System.AggregateException: One or more errors occurred. (Unable to write data to the transport connection: An established connection was aborted by the software in your host machine..)
 ---> System.IO.IOException: Unable to write data to the transport connection: An established connection was aborted by the software in your host machine..
 ---> System.Net.Sockets.SocketException (10053): An established connection was aborted by the software in your host machine.
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.CreateException(SocketError error, Boolean forAsyncThrow)
```

Logs on server side that indicate this issue:
```text
2022-07-29 09:15:40.104081+02:00 [warning] <0.8561.85> Giving credit to unknown subscription: 0
2022-07-29 09:15:43.106428+02:00 [warning] <0.8561.85> Giving credit to unknown subscription: 0
2022-07-29 09:16:11.714611+02:00 [warning] <0.8633.85> Giving credit to unknown subscription: 0
2022-07-29 09:16:14.716182+02:00 [warning] <0.8633.85> Giving credit to unknown subscription: 0
```

Gsantomaggio:
closed via https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/159

PR #159, opened 2022-08-26
Handle CreditResponse when subscription is unknown on server

jonnepmyra:
When calling Close() on a consumer while still handling messages in the consumer messagehandler, the Close() will throw exception, see issue #154 and the Close finally timeouts like:

```
System.TimeoutException: The operation has timed out.
   at RabbitMQ.Stream.Client.ManualResetValueTaskSource`1.System.Threading.Tasks.Sources.IValueTaskSource<T>.GetResult(Int16 token)
   at RabbitMQ.Stream.Client.Client.Request[TIn,TOut](Func`2 request, Nullable`1 timeout)
   at RabbitMQ.Stream.Client.Client.Request[TIn,TOut](Func`2 request, Nullable`1 timeout)
   at RabbitMQ.Stream.Client.Client.Close(String reason)
   at RabbitMQ.Stream.Client.Client.MaybeClose(String reason)
   at RabbitMQ.Stream.Client.Consumer.Close()
```

This PR aims to handle the [CreditResponse that can be issued by the server when the subscription is unknown ](https://github.com/rabbitmq/rabbitmq-server/blob/v3.10.x/deps/rabbitmq_stream/docs/PROTOCOL.adoc#credit) to the server.


Gsantomaggio:
@jonnepmyra thank you for your contribution. 
I changed the message log. It is enough to log the message without checking if the Consumer exists or not.

Gsantomaggio (approved):

Gsantomaggio:
cc @ricsiLT

ricsiLT:
Neat! Want a test drive before merging? I can certainly arrange that 😏

Also, is this version-specific from RMQ side? Or can I test on whatever 3.10.x?

Gsantomaggio:
>Or can I test on whatever 3.10.x?

yes you can use `3.10.x` Thank you

ricsiLT:
Works like charm!

Gsantomaggio:
Thank you all

Issue #165, opened 2022-09-20
Change Max len from Int to in64


Gsantomaggio:
```csharp
await streamSystem.CreateStream(new StreamSpec(StreamName)
        {
            MaxLengthBytes = 225485783,
            
        });
```

`MaxLengthBytes` is `int` it should be an `int64`

PR #166, opened 2022-09-21
Change definition of MaxLengthBytes stream specification - #165

PaoloLaurenti:
Change definition of MaxLengthBytes stream specification from int to ulong. 
I used unsigned type because negative values don't make sense for this setting.

Fixes: #165


Gsantomaggio (approved):

Gsantomaggio:
Thanks @PaoloLaurenti

PR #168, opened 2022-09-21
Producer super stream

### LINKED ISSUE #167: Super Stream Feature
Implement the client-side [super stream feature](https://blog.rabbitmq.com/posts/2022/07/rabbitmq-3-11-feature-preview-super-streams) 

The client needs a refractory. We need to implement an `IProducer` interface to implement:
- Producer 
- super-stream Producer

Per conversation with @lukebakken

Gsantomaggio:
Closes #167 

Based on `single-active consumer` branch. 
So we have to validate  https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/152 and than this PR.


Gsantomaggio:
How to test:
---
- Create super stream
```bash
rabbitmq-streams add_super_stream invoices --partitions 3
```

- Send messages

```csharp
 var config = new StreamSystemConfig();
 var system =  await StreamSystem.Create(config);
 var streamProducer = await system.CreateSuperStreamProducer(
       new SuperStreamProducerConfig()
       {
                SuperStream = "invoices",
                Routing = message => message.Properties.MessageId.ToString(),
                Reference = "ref",
        });
        for (ulong i = 0; i < 100_000; i++)
        {
            var message = new Message(Encoding.Default.GetBytes("hello"))
            {
                Properties = new Properties() {MessageId = $"hello{i}"}
            };
            await streamProducer.Send(i, message);
            if (i % 10_000 == 0)
            {
                Console.WriteLine(i);
            }
        }
    }
```

- Result should be:



super-stream

RabbitMQ.Stream.Client/StreamSystem.cs
```diff
@@ -120,7 +120,48 @@ private static void CheckLeader(StreamInfo metaStreamInfo)
...
+            {
+                throw new CreateProducerException(
+                    $"Batch Size must be bigger than 0");
+            }
+
+            if (superStreamProducerConfig.Routing == null)
+            {
+                throw new CreateProducerException(
+                    $"Routing Key Extractor must be provided");
+            }
```
Zerpet, line 141: Would make sense to make this validation in the `SuperStreamProducerConfig` constructor?
The rationale is to not let the user create an invalid `SuperStreamProducerConfig` that we know is not going to work; instead of letting create an invalid config and then failing to create the producer.

Zerpet (approved):

RabbitMQ.Stream.Client/StreamSystem.cs
Gsantomaggio: yes, good point. 
We should do the same to the standard producer

Gsantomaggio: @Zerpet, this would introduce a breaking change. The `SuperStreamProducerConfig`  would be:
`new SuperStreamProducerConfig(streamName)` instead of `new SuperStreamProducerConfig()`

So the `stream name` would be mandatory

Zerpet: Agreed with Gabriele to address this conversation in a follow up issue.

PR #169, opened 2022-09-22
Add client cert example to readme

ricsiLT:
I assume this isn't that common but some people might want this example in the repo for quick reference.


Gsantomaggio:
Thank you

PR #181, opened 2022-11-14
GetBytes(string).Length -> GetByteCount(string) + stackalloc in ReadString

sungam3r:
From MS sources:

```csharp
        public virtual byte[] Encoding.GetBytes(string s)
        {
            if (s == null)
                throw new ArgumentNullException(nameof(s),
                    SR.ArgumentNull_String);

            int byteCount = GetByteCount(s);
            byte[] bytes = new byte[byteCount];
            int bytesReceived = GetBytes(s, 0, s.Length, bytes, 0);
            Debug.Assert(byteCount == bytesReceived);
            return bytes;
        }
```

So `GetBytes(string).Length` call is pointless.


Gsantomaggio (approved):
That's interesting. Thank you

PR #204, opened 2022-12-21
Docker swarm superstream example

RossBugginsNHS:


Gsantomaggio:
@RossBugginsNHS Thank you, the example is great, but it is a bit out of scope with analytics, swagger etc. 
It will be difficult for us to maintain the example.

We have two ways at this point:
- Reduce the scope of the Example _only_ for super stream consumer/producer 
- you can post everything in your [repository](https://github.com/RossBugginsNHS/docker-rabbitmq-superstreams)  and we will be glad to link your repo in the example section

RossBugginsNHS:
No probs.  I needed that as a way of showing what's actually happening after changing number of consumers, so will leave in for now. I've put it at https://github.com/RossBugginsNHS/swarm-rabbitmq-superstreams cheers, Ross

Issue #205, opened 2023-01-03
Unclosed connections (memoryleak)


jonnepmyra:
Hi!

We're using streams with shortlived stream-consumers.

Our applications connect at some offset in the stream and read a couple of messages until it closes().

It's not uncommon to see a warning, followed by an error in the logs under the conditions mentioned above. The logs look like:

```
09:25:30.305 [WARN] Received credit response for unknown subscription 0, ResponseCode SubscriptionIdDoesNotExist
09:25:30.300 [ERROR] Error removing the consumer id: 0 from the server
System.TimeoutException: The operation has timed out.
   at RabbitMQ.Stream.Client.RawConsumer.Close()
```

Each of these logged errors, is always followed by a connection not being closed towards RabbitMQ.

The consequence is that the memory footprint is increased in both the client application and on the RabbitMQ Server.

Gsantomaggio:
Thank you for reporting.

Is the bug reproducible in somehow?

Can you please provide a snippet or GitHub repo with the bug so we can test it?

jonnepmyra:
Hi, Thanks for the quick response...

I'll see if I can reproduce in a simple way...

In the meantime, I've added some more debug log to my local fork and figured out that when the unsubscribe command times out and the code continues to close the client,  `conumers == 1` at this location:

https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/ce2e8475cb2c8e916fc3f8fc370b30c1e69b6eae/RabbitMQ.Stream.Client/Client.cs#L575-L578

which could be the reason that the connection is never closed?

Gsantomaggio:
Yes, it could be. 
this error:
`09:25:30.305 [WARN] Received credit response for unknown subscription 0, ResponseCode SubscriptionIdDoesNotExist` means that the subscriber does not exist anymore on the server, but the client sends the credit.

It could happen when the consumer is closing when it still receiving the messages from the stream

jonnepmyra:
Seems like there's some sort of deadlock and the `.GetAwaiter().GetResult()` in the MessageDispatcher which was added in #184 is causing the problem....

If the MessageDispatcher is "reset" to pre-rc8, closing consumers works like a charm.

This "test" could be used to reproduce... (strip the logging part if you want, or add the following nuget packages to get it working)

```
    <PackageReference Include="Divergic.Logging.Xunit" Version="4.2.0" />
[...]

PR #184, opened 2022-11-22
Add GetAwaiter().GetResult(); on MessageHandler

Gsantomaggio:
Small improvement, the `GetAwaiter().GetResult();` will wait for the complete task. it takes the memory under control.

Signed-off-by: Gabriele Santomaggio <G.santomaggio@gmail.com>

cc @ricardSiliuk @ricsiLT

PR #206, opened 2023-01-03
Add cancellation token during subscribe

Gsantomaggio:
Fixes https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/205

In this way, the thread is not locked, and the:
```csharp
new MessageContext(message.MessageOffset, TimeSpan.FromMilliseconds(chunk.Timestamp)),
                            message).Wait(_token);
```

is in wait(...), so there is not the thread pool exhaustion.

@ricardSiliuk @ricsiLT it should not change anything. Please feel free to add any comments.

@jonnepmyra  do you have a chance to test it? 
 
### DIFF
--- a/RabbitMQ.Stream.Client/RawConsumer.cs
+++ b/RabbitMQ.Stream.Client/RawConsumer.cs
@@ -316,12 +316,10 @@ namespace RabbitMQ.Stream.Client
 
             var closed = _client.MaybeClose($"_client-close-subscriber: {_subscriberId}");
             ClientExceptions.MaybeThrowException(closed.ResponseCode, $"_client-close-subscriber: {_subscriberId}");
-            _disposed = true;
             _logger.LogDebug("Consumer {SubscriberId} closed", _subscriberId);
             return result;
         }
 
-        //
         private void Dispose(bool disposing)
         {
             if (!disposing)
@@ -334,11 +332,18 @@ namespace RabbitMQ.Stream.Client
                 return;
             }
 
-            var closeConsumer = Close();
-            closeConsumer.Wait(TimeSpan.FromSeconds(1));
-            ClientExceptions.MaybeThrowException(closeConsumer.Result,
-                $"Error during remove producer. Subscriber: {_subscriberId}");
-            _cancelTokenSource.Dispose();
+            try
+            {
+                var closeConsumer = Close();
+                closeConsumer.Wait(TimeSpan.FromSeconds(1));
+                ClientExceptions.MaybeThrowException(closeConsumer.Result,
+                    $"Error during remove producer. Subscriber: {_subscriberId}");
+            }
+            finally
+            {
+                _cancelTokenSource.Dispose();
+                _disposed = true;
+            }
         }
 
         public void Dispose()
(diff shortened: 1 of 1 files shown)


jonnepmyra:
Hi!

Thank your for the quick response and patch @Gsantomaggio 

We've been testing it during the day with various workloads and it's definitely alot better!

At first we could still se about 0.5% of all closed connections get deadlocked with an error:
`Error removing the consumer id: 0 from the server
`

... but after tweaking our consumer messagehandlers that contained TaskCompletionSources with `TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);` it works better and we're no longer able to reporduce the problem!!

The reason why I'm mentioning the TaskCompletionSource tweak in our application code is that even if it works well now, and we're super grateful for this being fixed, the code still seems to be able to deadlock under certain conditions but I might be wrong... 

The Wait() in:
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/9d341701b5d3e55e0d9a1d791e3aa0bc88196c8c/RabbitMQ.Stream.Client/RawConsumer.cs#L160-L162



feels so wrong (as mentioned before in https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/106#issuecomment-1240246453 ) but on the other hand, I've no better suggestion at the very moment if `WaitAsync()` leads to thread stravation.

Thanks again!

Gsantomaggio:
@jonnepmyra let me clarify some points:

- We can't use `WaitAsync` since we will end up on https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/184#issuecomment-1326154401


- We can't use aync Tasks here: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/9d341701b5d3e55e0d9a1d791e3aa0bc88196c8c/RabbitMQ.Stream.Client/RawConsumer.cs#L160-L162
Because with `ref SequenceReader`, it is not possible:

```
Parameters of type `SequenceReader<byte>` cannot be declared in async local functions
```

- We introduced `SequenceReader<byte>` for performance problems here: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/180

- `.Wait(_token); ` should not go in dead-lock (unless there is some blocking code in your consumer).


Have questions:

- > code still seems to be able to deadlock
     - Have you tried without your consumer code to reproduce the lock?
     - What do you mean by deadlock? Do you still have a memory leak? 
     - Have you tried the last commit https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/206/commits/9d341701b5d3e55e0d9a1d791e3aa0bc88196c8c ?
     
Thank you for trying the PR.  :)

p.s. If you want, you can find me at https://rabbitmq-slack.herokuapp.com/

jonnepmyra:
Hi!

Thanks for taking time to explain.

>   * Have you tried without your consumer code to reproduce the lock?

Yes, and I have not been able to reproduce it.

>   * What do you mean by deadlock? Do you still have a memory leak?

The Memoryleak seems to be fixed now, no connections left unclosed after tweaking application specific consumer code!

>   * Have you tried the last commit [9d34170](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/commit/9d341701b5d3e55e0d9a1d791e3aa0bc88196c8c) ?

Yes, it seems fine

Cheers!

Gsantomaggio:
Great! We will merge it soon!
Ty

Gsantomaggio:
Let's wait for some feedback from @ricardSiliuk @ricsiLT then, we can merge and release.

PR #217, opened 2023-01-18
Fix the read in the AMQP parser

Gsantomaggio:
- Change the string read: use the offset len from the protocol instead
  of encoding.GetByteCount(value). In some case the message properties can't be UFT8.
- Use Encoding.ASCII.GetString in case of FormatCode.Sym8, that's more correct.
- Add the AMQPlite dependency on the Test package to simulate the use case where the
  parser failed.

The issue was reported by @thorhalvor 

To stimulate the issue ,you need to send a message using the [AMQP 1.0 client](https://github.com/Azure/amqpnetlite):

```csharp

var address = new Address("amqp://guest:guest@localhost:5672");
var connection = new Connection(address);
var session = new Session(connection);

Message message = new Message("A");
message.Properties = new Properties()
{
    MessageId = "1",
    Subject = "test",
    ContentType = "text/plain"
};
message.ApplicationProperties = new ApplicationProperties()
{
    Map = { { "key1", "value1" }, { "key2", 2 } }
};

// stream queue
var sender = new SenderLink(session, "mixing", "/amq/queue/my_stream_queue");
sender.Send(message);
```

and read it with the stream client; you will receive:
```shell
=============================================
fail: RabbitMQ.Stream.Client.Reliable.Consumer[0]
      Unexpected error while parsing message. Message will be skipped. Please report this issue to the RabbitMQ team on GitHub https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/is
sues
      System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'dataCode: 0 not handled')
         at RabbitMQ.Stream.Client.RawConsumer.<ParseChunk>g__DispatchMessage|10_0(SequenceReader`1& sequenceReader, UInt64 i, <>c__DisplayClass10_0& )
```

Try to read again with this PR, and it works!

@thorhalvor [already tested](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/217#issuecomment-1387221891) 

### DIFF
--- a/.ci/install.ps1
+++ b/.ci/install.ps1
@@ -158,4 +158,4 @@ Write-Host '[INFO] Getting RabbitMQ status...'
 
 $ErrorActionPreference = 'Continue'
 Write-Host '[INFO] Enabling plugins...'
-& $rabbitmq_plugins_path enable rabbitmq_management rabbitmq_stream rabbitmq_stream_management
+& $rabbitmq_plugins_path enable rabbitmq_management rabbitmq_stream rabbitmq_stream_management rabbitmq_amqp1_0
--- a/.ci/versions.json
+++ b/.ci/versions.json
@@ -1,4 +1,4 @@
 {
   "erlang": "25.2",
-  "rabbitmq": "3.11.6"
+  "rabbitmq": "3.11.7"
 }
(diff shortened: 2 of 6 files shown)


thorhalvor:
We had a problem consuming messages that was published with AmqpNetLite (Amqp 1.0) library. Can confirm that 
this commit will fix that problem for us and messages can be consumed without errors. Well Done!

thorhalvor:
Nice work! This fix the error we got when message was published with AmqpNetLite (amqp1.0)

lukebakken (approved):

Issue #145, opened 2022-07-04
Negotiate frame_max 


Gsantomaggio:
In the next RabbitMQ [version will be possible](https://github.com/rabbitmq/rabbitmq-server/pull/5131) to change the frame_size for the stream. 

The client can ask to change the frame_size, but it can't be bigger than the server. 

At the moment there are no controls

PR #220, opened 2023-01-26
Handle System.InvalidOperationException in case of big body buffer send

Gsantomaggio:
Fixes https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/220 and part of https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/145

How to reproduce the issue:
---
See https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/221#issuecomment-1408242039
Important:
 - Server has to be remote
 - the body needs to be bigger than 3k

Main changes:
---

- Make the `WriteCommand` Async:
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/5c695ef3862dd2f6ec0b72c52570e9c282a0ef34/RabbitMQ.Stream.Client/Connection.cs#L99-L114

Remove sync Waits (because of https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2012) :
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/5c695ef3862dd2f6ec0b72c52570e9c282a0ef34/RabbitMQ.Stream.Client/Connection.cs#L89-L97
 

- Make the Async Blocking:
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/5c695ef3862dd2f6ec0b72c52570e9c282a0ef34/RabbitMQ.Stream.Client/RawProducer.cs#L161-L164
It caused a lot of timeout operations but without any action to handle the problem.

- Move the semaphore scope on the Producer Class:
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/5c695ef3862dd2f6ec0b72c52570e9c282a0ef34/RabbitMQ.Stream.Client/Reliable/Producer.cs#L207

It caused `Client timeout Error` in case of multi-threads publish

### DIFF
--- a/RabbitMQ.Stream.Client/Client.cs
+++ b/RabbitMQ.Stream.Client/Client.cs
@@ -239,15 +239,11 @@ namespace RabbitMQ.Stream.Client
 
         public async ValueTask<bool> Publish(Publish publishMsg)
         {
-            var publishTask = Publish<Publish>(publishMsg);
-            if (!publishTask.IsCompletedSuccessfully)
-            {
-                await publishTask.ConfigureAwait(false);
-            }
+            var publishTask = await Publish<Publish>(publishMsg);
 
             publishCommandsSent += 1;
             messagesSent += publishMsg.MessageCount;
-            return publishTask.Result;
+            return publishTask;
         }
 
         public ValueTask<bool> Publish<T>(T msg) where T : struct, ICommand
--- a/RabbitMQ.Stream.Client/Connection.cs
+++ b/RabbitMQ.Stream.Client/Connection.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.IO.Pipelines;
 using System.Net;
 using System.Net.Sockets;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace RabbitMQ.Stream.Client
@@ -20,7 +21,7 @@ namespace RabbitMQ.Stream.Client
         private Func<Memory<byte>, Task> commandCallback;
         private readonly Func<string, Task> closedCallback;
         private int numFrames;
-        private readonly object writeLock = new();
+        private readonly SemaphoreSlim writeLock = new SemaphoreSlim(1, 1);
         internal int NumFrames => numFrames;
         private bool isClosed = false;
 
@@ -87,8 +88,7 @@ namespace RabbitMQ.Stream.Client
 
         public async ValueTask<bool> Write<T>(T command) where T : struct, ICommand
         {
-
-            var flushTask = await WriteCommand(command);
+            await WriteCommand(command);
 
             // Let's check if this is completed synchronously before invoking the async state machine
             // if (!flushTask.IsCompletedSuccessfully)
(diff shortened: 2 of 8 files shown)


RabbitMQ.Stream.Client/Client.cs
```diff
@@ -239,15 +239,11 @@ public static async Task<Client> Create(ClientParameters parameters, ILogger log
 
         public async ValueTask<bool> Publish(Publish publishMsg)
         {
-            var publishTask = Publish<Publish>(publishMsg);
-            if (!publishTask.IsCompletedSuccessfully)
```
Gsantomaggio, line 243: For more info about this change, see:
 https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2012

RabbitMQ.Stream.Client/Connection.cs
```diff
@@ -85,41 +86,28 @@ private static System.IO.Stream MaybeTcpUpgrade(NetworkStream networkStream, Ssl
...
-            }
-
-            return flushTask.Result.IsCompleted;
+            await WriteCommand(command);
+            // we return true to indicate that the command was written
+            return true;
         }
 
-        private void WriteCommand<T>(T command) where T : struct, ICommand
+        private async Task WriteCommand<T>(T command) where T : struct, ICommand
```
Gsantomaggio, line 96: This is the main change. 
It makes `WriteCommand ` async using `WriteAsync` Add `FlushAsync`  instead of [sync wait the task](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/220/files#diff-71215640c5968601ad0a124946241692beb3c931cc450bb43bb7061e8f84c47eL245)

RabbitMQ.Stream.Client/RawProducer.cs
```diff
@@ -197,26 +197,9 @@ internal void PreValidateBatch(List<(ulong, Message)> messages)
...
-        private async Task SemaphoreWait()
-        {
-            if (!await _semaphore.WaitAsync(0) && !_client.IsClosed)
-            {
-                // Nope, we have maxed our In-Flight messages, let's asynchronously wait for confirms
-                if (!await _semaphore.WaitAsync(TimeSpan.FromSeconds(1)).ConfigureAwait(false))
-                {
-                    _logger.LogWarning("Semaphore Wait timeout during publishing.");
-                }
-            }
```
Gsantomaggio, line 209: Remove this Async Await. In some case, it raises `semaphorefullexception` 
after:
```
_logger.LogWarning("Semaphore Wait timeout during publishing.");
```
the code continue to work failing the scope of the `lock`

RabbitMQ.Stream.Client/RawProducer.cs
```diff
@@ -268,13 +256,13 @@ private async Task ProcessBuffer()
                         messages.Add((msg.PublishingId, msg.Data));
                         if (messages.Count == _config.MessagesBufferSize)
                         {
-                            await SendMessages(messages).ConfigureAwait(false);
+                            await SendMessages(messages);
```
Gsantomaggio, line 259: It is not needed due of `await`

RabbitMQ.Stream.Client/Connection.cs
```diff
@@ -85,41 +86,28 @@ private static System.IO.Stream MaybeTcpUpgrade(NetworkStream networkStream, Ssl
...
         {
             // Only one thread should be able to write to the output pipeline at a time.
-            lock (writeLock)
+            await writeLock.WaitAsync();
```
Gsantomaggio, line 99: Remove the lock and use a `semaphore(1,1)` to simulate the lock in async method

RabbitMQ.Stream.Client/Connection.cs
```diff
@@ -85,41 +86,28 @@ private static System.IO.Stream MaybeTcpUpgrade(NetworkStream networkStream, Ssl
...
-            var flushTask = FlushAsync();
-
-            // Let's check if this is completed synchronously before invoking the async state machine
-            if (!flushTask.IsCompletedSuccessfully)
-            {
-                await flushTask.ConfigureAwait(false);
-            }
-
-            return flushTask.Result.IsCompleted;
+            await WriteCommand(command);
```
Gsantomaggio, line 91: Now everything is async with the `await` we `return true;` to leave the API compatibility

RabbitMQ.Stream.Client/RawProducer.cs
```diff
@@ -150,17 +150,19 @@ public async ValueTask Send(ulong publishingId, List<Message> subEntryMessages,
         {
             if (subEntryMessages.Count != 0)
             {
-                await SemaphoreWait();
+                await SemaphoreAwaitAsync();
                 var publishTask =
                     _client.Publish(new SubEntryPublish(_publisherId, publishingId,
                         CompressionHelper.Compress(subEntryMessages, compressionType)));
-                if (!publishTask.IsCompletedSuccessfully)
```
Gsantomaggio, line 157: Changed due of https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2012

Zerpet (dismissed):

lukebakken (changes requested):
As far as I know, since this is a general-purpose library, every `await` should have `ConfigureAwait(false)` - https://devblogs.microsoft.com/dotnet/configureawait-faq/#when-should-i-use-configureawaitfalse

PR #230, opened 2023-02-03
Improve the reconnect

Gsantomaggio:
Handle some edge cases where the producer or consumer could go into a deadlock waiting for the semaphore.

Add the Cancellation token producer side and call the token cancellation when the producer and the consumer are forced to disconnect.

I did this [CaosTest](https://github.com/Gsantomaggio/rabbitmq-utils/tree/master/client_caos/Caos) for the stream client.

Where the connections are closed randomly.
```
Messages sent: 344800 -Messages confirmed: 227737 - Messages error: 1024 - Total 228761  Messages consumed: 169337
Messages sent: 397000 -Messages confirmed: 279919 - Messages error: 116082 - Total 396001  Messages consumed: 206422
```

Client should be able to reconnect and continue to produce and consume 

 

### DIFF
--- a/RabbitMQ.Stream.Client/Client.cs
+++ b/RabbitMQ.Stream.Client/Client.cs
@@ -190,7 +190,8 @@ namespace RabbitMQ.Stream.Client
         {
             if (ConnectionClosed != null)
             {
-                await ConnectionClosed?.Invoke(reason)!;
+                var t = ConnectionClosed?.Invoke(reason)!;
+                await t.ConfigureAwait(false);
             }
         }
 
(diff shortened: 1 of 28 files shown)


Gsantomaggio:
cc @iuribrindeiro @ricardSiliuk @ricsiLT

RabbitMQ.Stream.Client/Client.cs
```diff
@@ -547,30 +547,38 @@ public async Task<CloseResponse> Close(string reason)
...
-            return result;
+            return new CloseResponse(0, ResponseCode.Ok);
+
         }
 
         // Safe close 
         // the client can be closed only if the publishers are == 0
         // not a public method used internally by producers and consumers
-        internal CloseResponse MaybeClose(string reason)
+        internal async Task<CloseResponse> MaybeClose(string reason)
```
iuribrindeiro, line 577: ⭐

RabbitMQ.Stream.Client/Client.cs
```diff
@@ -547,30 +547,38 @@ public async Task<CloseResponse> Close(string reason)
...
         {
             if (publishers.Count == 0 && consumers.Count == 0)
             {
-                return Close(reason).Result;
+                await Close(reason).ConfigureAwait(false);
```
iuribrindeiro, line 581: <3

iuribrindeiro:
> cc @iuribrindeiro @ricardSiliuk @ricsiLT

@Gsantomaggio please see: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/231

ricardSiliuk:
Should this token stay `private`/`protected` or be made public? Just thinking "out loud" but I'd probably want to know if consumer/producer was somehow unexpectedly cancelled by linking new CTS with your `Token`?

Though if the only situation where this happens is when end user calls `Close()`, maybe he doesn't need to know about your CTS... not sure.

Gsantomaggio:
Thank you @iuribrindeiro @lukebakken @ricardSiliuk

PR #232, opened 2023-02-06
Enable possibility for RawConsumers to use SuperStreams

jonnepmyra:
Had a quick chat with @Gsantomaggio about this..

This opens up the possibility to use a `RawConsumer` (with all it options) with SuperStream capabilities which has advantages in some edge-case-scenarios like:

* You need to start a consumer for a single partition, and not all partitions of the superstream
* You need specific event-callbacks for each partition in the SuperStream
* You want to implement your own `ConnectionClosedHandler` and/or `MetadataHandler`


Gsantomaggio:
Closed in favour of  https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/250

PR #233, opened 2023-02-08
Small Fix read boolean

Gsantomaggio:
Fix the read boolean. The offeset is moved corretly. Add test for this use case.


lukebakken (commented):
👍

Issue #227, opened 2023-02-02
Implement DeduplicationProducer Class


Gsantomaggio:
The deduplication feature is only available on the `RawProducer` 
The idea is to have an intelligent layer (as the `Producer`) to handle the deduplication.

 
The new class `DeduplicationProducer` will have only the methods to handle the deduplication. It will be easier for the user to handle this use case.

In the `Producer` class, the `Reference` field will be removed since it is needed only for deduplication. 
In the `DeduplicationProducer` class the the `Reference` field will be mandatory

Gsantomaggio:
Per conversation with @Zerpet we decide to mark deprecated the `Reference` in the  `Producer` and explain why in the doc

PR #234, opened 2023-02-09
Implement Deduplicating Producer

Gsantomaggio:
Fixes https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/227

Implement the `DeduplicationProducer` make the `ProducerConfig.Reference` deprecated.
The `Producer` does not need the `Reference` to work. `Reference` is needed only for the Deduplication part.

The low-level Class is the `RawProducer` class; this class sets the correct parameters to enable the deduplication and to make it easy to use. 

 

### DIFF
--- a/RabbitMQ.Stream.Client/Reliable/Producer.cs
+++ b/RabbitMQ.Stream.Client/Reliable/Producer.cs
@@ -35,15 +35,15 @@ public record ProducerConfig : ReliableConfig
 
     /// <summary>
     /// Reference used for deduplication.
-    /// For the Producer Class it is not needed to set this value.
-    /// See DeduplicationProducer for Deduplication Messages where this value is needed.
+    /// For the Producer Class, it is not needed to set this value
+    /// See DeduplicatingProducer for Deduplication Messages where this value is needed.
     /// </summary>
     internal string _reference;
 
     public string Reference
     {
         get { return _reference; }
-        [Obsolete("Deprecated. Use ClientProvidedName instead. Se DeduplicationProducer for Deduplication Messages ",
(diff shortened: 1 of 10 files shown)


michaelklishin:
Should this be `DeduplicatingProducer`, a producer that deduplicates messages vs. a producer of deduplication? :) Sorry.

lukebakken (changes requested):
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/234#issuecomment-1426891808

Thanks!

RabbitMQ.Stream.Client/Reliable/Producer.cs
```diff
@@ -232,7 +268,7 @@ public async ValueTask Send(Message message)
         }
         finally
         {
-            SemaphoreSlim.Release();
+            // SemaphoreSlim.Release();
         }
```
Zerpet, line 272: If we are not locking the semaphore in this method, can we delete this `finally` block?

Tests/DeduplicationProducerTests.cs
```diff
@@ -0,0 +1,131 @@
...
+
+/* Unmerged change from project 'Tests(net7.0)'
+Before:
+using System.Collections.Generic;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+After:
+using System.Threading.Tasks;
+*/
```
Zerpet, line 13: I guess we can remove those? 🙈

Zerpet (dismissed):
Looking good. Left a few cosmetic suggestions, but up to you if you want to merge as-is.

Gsantomaggio:
Thanks @Zerpet fixed here https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/234/commits/1780afd2e044a06618e6625efb2433fe25d931d2

PR #236, opened 2023-02-16
Small super stream improvements

Gsantomaggio:
Check if the result is null. Add a log when the write fails. Improve the super stream example

### DIFF
--- a/Examples/SuperStream/SuperStreamConsumer.cs
+++ b/Examples/SuperStream/SuperStreamConsumer.cs
@@ -9,14 +9,10 @@ using RabbitMQ.Stream.Client.Reliable;
 
 namespace SuperStream;
 
-public static class SuperStreamConsumer
+public class SuperStreamConsumer
 {
     public static async Task Start(string consumerName)
     {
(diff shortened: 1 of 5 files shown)


Examples/SuperStream/SuperStreamConsumer.cs
```diff
@@ -26,13 +35,20 @@ public static async Task Start(string consumerName)
             // must have the same ReferenceName for all the consumers
             Reference = "MyApp",
             OffsetSpec = new OffsetTypeFirst(),
-            
+            ConsumerUpdateListener = async (reference, stream, isActive) =>
+            {
+                Console.WriteLine($"******************************************************");
+                Console.WriteLine($"reference {reference} stream {stream} is active: {isActive}");
+                Console.WriteLine($"******************************************************");
```
Zerpet, line 42: Since we created a `logger`, shall we log everything using the `logger` ?

Zerpet (commented):
Early feedback 🐦

Zerpet (approved):

Issue #246, opened 2023-03-03
Single Active consumer blocked in `Waiting Status`


Gsantomaggio:
Use case:
====
- SuperStream with four partitions.
- Single active consumer active.
- More services start at the same time.
- The consumer sometimes takes a couple of ms to several seconds to handle a single message.

Symptom
====
- Sometimes when the consumers start, they go into `waiting` status,apparently without exiting.

Problem:
====

- When a consumer (A) starts requesting ten credits. 
- The consumer needs to handle all these chunks by blocking the socket thread.
- When the second consumer (B) starts, the server sends the `ConsumerUpdateListener` to  (A) and (B), but the (A) socket is blocked to handle the chunks. So the `ConsumerUpdateListener` won't be called. 
- As a side effect: (A) is blocked from handling the chunks and the server can't promote (B) as next single active consumer
See: 
 ![Screenshot 2023-03-03 at 12 32 25](https://user-images.githubusercontent.com/386987/222714194-b2218dc0-95db-4956-b193-3e2d806b8523.png)

How to reproduce:
===

Run more instances of this: 
https://github.com/Gsantomaggio/rabbitmq-utils/blob/master/client_caos/Caos/SuperStreamRaw.cs

Reported by @jonnepmyra

PR #250, opened 2023-03-09
Run the  Consumer MessageHandler in a Task

Gsantomaggio:
Thanks to @Zerpet for help and to @jonnepmyra for rasing the issue and testing the pr 
Fixes https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/246

Run the Consumer MessageHandler in a Task
---
The chunks are now processed in a separate [Task](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/250/files#diff-b08ff9a3d581af9df4f7e209452830c93f2367debb487e937e158870332fbd1eR263-R272)
In this way, the socket thread is free to handle other requests like `ConsumerUpdate`

Convert `DispatchMessage` method from `void` to `async Task`
---

[`DispatchMessage`](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/250/files#diff-b08ff9a3d581af9df4f7e209452830c93f2367debb487e937e158870332fbd1eR195-R198) is now an `async task` by removing the `wait` the code is aync


lukebakken (commented):
I've reviewed once but want to re-read again!

Gsantomaggio:
Thank you, @lukebakken. I convert the pr into a draft. I am working with @jonnepmyra to validate it; it has yet to be ready.

jonnepmyra:
Been testing this in many different scenarios the last week, and the current state of this PR is a **huge improvement** that solves multiple problems that've experienced when using SuperStreams.

Some examples of problems that's been solved:
* Deadlocks for slow consumers
* Running out of credits on multiple "single active consumer" switches
* None of the consumers where activated (== all consumers is "stuck" in a "waiting" status)
* Consumer connections that were silently dropped

We're still struggling with cases of "all consumers are stuck in waiting status", but this happends alot less frequently with this PR, and seems to correlate with heavy load in RabbitMQ (or possibly the host of the application instance) and more investigation is needed (Will continue a separate discussion with @Gsantomaggio about this, since it seems to be out of context of this PR and very hard to reproduce).

RabbitMQ.Stream.Client/Consts.cs
```diff
@@ -8,11 +8,23 @@ namespace RabbitMQ.Stream.Client
 {
     internal static class Consts
     {
+        private static readonly Random s_random = new();
```
ricardSiliuk, line 11: I think you can drop this and just use `Random.Shared.Next()` on L22 and L27 :)

https://andrewlock.net/building-a-thread-safe-random-implementation-for-dotnet-framework/

PR #259, opened 2023-04-07
Switch to Random.Shared

Pliner:
Hi,

`Random.Shared` is already used in other places in the codebase and I don't see any reasons why it should not be used here.


Gsantomaggio (approved):

Gsantomaggio:
Thank you @Pliner

Gsantomaggio (approved):

PR #260, opened 2023-04-07
Iterate _lastOffsetConsumed in a more simple way

Pliner:
Hi,

Just one more small addition.

It does the same job, but in a more simple and idiomatic way.


Gsantomaggio (approved):

Gsantomaggio:
thank you @Pliner

PR #265, opened 2023-04-18
Expose initial credits configuration

Gsantomaggio:
ref: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/discussions/264

### DIFF
--- a/RabbitMQ.Stream.Client/IConsumer.cs
+++ b/RabbitMQ.Stream.Client/IConsumer.cs
@@ -16,6 +16,8 @@ public interface IConsumer
 
 public record IConsumerConfig : INamedEntity
 {
+    private ushort _initialCredits = 2;
+
     // StoredOffsetSpec configuration it is needed to keep the offset spec.
     // since the offset can be decided from the ConsumerConfig.OffsetSpec.
     // and from ConsumerConfig.ConsumerUpdateListener.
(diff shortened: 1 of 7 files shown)


Gsantomaggio:
@debarisi Not sure this will help. Have you tried it? 

The following diagram explains how the flow works. The new credit will be appended to the buffer chunk.


chunk_flow drawio


Are you able to trace `Channel.CreateBounded` status?
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/60dfb718323e5ee9e8b20aff5f0bdb9983ffe0a2/RabbitMQ.Stream.Client/RawConsumer.cs#L115

debarisi:
@Gsantomaggio, you are right. My apologies. 
Even in the documentation, it clearly says:
_A credit represents a chunk of messages that the broker is allowed to send to the consumer._
In my tests, I asked for a credit on every message, which was why I had a bad performance.

Thanks for the clarification.

RabbitMQ.Stream.Client/IConsumer.cs
```diff
@@ -40,4 +42,24 @@ public record IConsumerConfig : INamedEntity
...
+    // The RawConsumer uses this value to create the Channel buffer so all the chunks will be stored in the buffer memory.
+    // The default value it is usually a good value.
+    public ushort InitialCredits
+    {
+        get => _initialCredits;
+        set
+        {
+            if (value < 1)
+            {
+                throw new ArgumentException("InitialCredits must be greater than 0. Default value is 2.");
```
ricsiLT, line 59: Maybe make this message dynamic and reference magic value? Might get out of sync in case you'll want to increase this in the future

Gsantomaggio: Done via https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/265/commits/10ebdec75c2293490740f4805ac09cd40711d212

Gsantomaggio:
Per discussion with @Zerpet we will expose this parameter.

Zerpet (approved):

Zerpet:
I guess we can ship a new minor for this, no need to wait for 2.0.0. What do you think?

Gsantomaggio:
>I guess we can ship a new minor for this, no need to wait for 2.0.0. What do you think?

yes agree 100%!

PR #272, opened 2023-06-07
Fix RawConsumer bug in DEBUG log statement causing exceptions

ryan-ingotsoft:
Bug in DEBUG level log statement for creation of `RawConsumer`. Occurs only when DEBUG level was specified for the consumer. Confirmed using the following:

```
var factory = LoggerFactory.Create(builder =>
            {
                builder.AddSimpleConsole();
                builder.AddFilter("RabbitMQ.Stream", LogLevel.Debug);
            });

...
 var consumerLogger = factory.CreateLogger<Consumer>();
 var consumer = await system.CreateRawConsumer(
                new RawConsumerConfig(stream)
                {
                    Reference = "consumer",
                    MessageHandler = async (consumer, ctx, message) =>
                    {
                        await Task.CompletedTask;
                    }
                }, consumerLogger);
```

### DIFF
--- a/RabbitMQ.Stream.Client/RawConsumer.cs
+++ b/RabbitMQ.Stream.Client/RawConsumer.cs
@@ -105,7 +105,7 @@ namespace RabbitMQ.Stream.Client
             _logger = logger ?? NullLogger.Instance;
             _initialCredits = config.InitialCredits;
             _logger.LogDebug("creating consumer {Consumer} with initial credits {InitialCredits}, " +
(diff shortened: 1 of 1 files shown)


Gsantomaggio (changes requested):

Gsantomaggio (approved):

Gsantomaggio (approved):

Gsantomaggio:
Thank you

PR #276, opened 2023-06-13
Implement Chunk Filter 

Gsantomaggio:
The PR implements the chunk filter feature, see https://github.com/rabbitmq/rabbitmq-server/pull/8207
Check if the broker supports the broker filter.
Check if the broker is `>= 3.11.0` to validate the exchange version 

How to test:
---
```
  docker run -it --rm --name rabbitmq -p 5552:5552 -p 15672:15672 \
    -e RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS='-rabbitmq_stream advertised_host localhost' \
    rabbitmq:3.13.0-beta.2-management
```

Then:
```
docker exec rabbitmq    rabbitmq-plugins enable rabbitmq_stream_management
```

A full example:  `docs/StreamFilter/Filter`


RabbitMQ.Stream.Client/PublishFilter.cs
```diff
@@ -0,0 +1,132 @@
...
+                }
+                catch (Exception e)
+                {
+                    // If there is an error on _filterValueExtractor we skip the message.
+                    // If there is an error on _filterValueExtractor the buffer here is still consistent.
+                    // so we can skip and continue
+                    _logger.LogError(e, "Error writing the filter message. Message won't be sent"
+                                        + "Check the filter value function");
+                }
+            }
```
Zerpet, line 112: I _think_ the write function may not behave correctly in certain error scenarios. For example, it always writes the number of messages (let's say N number of messages). However, if there's an exception in `_filterValueExtractor()`, the message will be skipped. If we skip M number of messages, we will be sending `N-M` messages, however, we wrote that this frame has `N` messages. I'm not sure what's going to happen in the server with this frame 🤔 

I haven't checked what the Java implementation does, but I would consider re-throwing and logging an error somewhere, probably in the caller of `write`

RabbitMQ.Stream.Client/RasSuperStreamConsumer.cs
Zerpet, line 1: Shouldn't the filename be `RawSuperStreamConsumer.cs` ?
I just noticed it is Ra **s** SuperStreamConsumer.cs

RabbitMQ.Stream.Client/RawProducer.cs
```diff
@@ -213,7 +224,17 @@ internal void PreValidateBatch(List<(ulong, Message)> messages)
...
+            switch (IsFilteringEnabled)
+            {
+                case true:
+                    await _client.Publish(new PublishFilter(_publisherId, messages, _config.Filter.FilterValue, _logger))
+                        .ConfigureAwait(false);
+                    break;
+                case false:
+                    await _client.Publish(new Publish(_publisherId, messages)).ConfigureAwait(false);
+                    break;
+            }
```
Zerpet, line 236: I'm ok with the `switch`, but doesn't read better using an `if-else` ?

Zerpet (commented):
I have to review the tests and the docs, but I have some initial comments to get going.

RabbitMQ.Stream.Client/PublishFilter.cs
Gsantomaggio: [Here](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/stream-filtering/RabbitMQ.Stream.Client/PublishFilter.cs#L36-L44) there is the fix for the message count. 

The size calculation happens before sending the messages so we can decrease the message count to send. 

Read the internal conversation about the [error handling](https://vmware.slack.com/archives/CSC92F50B/p1687180929189639)  in the case of Filter. The same problem happens when we consume the message for the post filter.  See [here](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/stream-filtering/RabbitMQ.Stream.Client/RawConsumer.cs#L251-L255)

docs/StreamFilter/StreamFilter/Start.cs
Zerpet, line 1: 🔥

kubernetes/stream_cluster.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
   namespace: stream-clients-test
 spec:
   replicas: 1
-  image: rabbitmq:3.11.11-management
+  image: rabbitmq:3.11.17-management
```
Zerpet, line 13: This is fine, although latest 3.11 is 3.11.20. Any reason to not bump to the latest? Or even to tag `rabbitmq:3.11-management` that tracks the latest patch?

Tests/FilterTest.cs
```diff
@@ -0,0 +1,271 @@
...
+        await c.Close();
+        await SystemUtils.CleanUpStreamSystem(system, stream).ConfigureAwait(false);
+    }
+
+    // This test is checking that the filter is working as expected
+    // We send 100 messages with two different states (Alabama and New York)
+    // By using the filter we should be able to consume only the messages from Alabama 
+    // and the server has to send only one chunk with all the messages from Alabama
+    [Fact]
+    public async void FilterShouldReturnOnlyOneChuck()
```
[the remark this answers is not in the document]
Gsantomaggio: OMG :)!

Zerpet (dismissed):

Zerpet:
It looks like this line is causing the CI failure:

https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/d9ab4f8ca9e56a41f20c239f0099d03a874245a5/.ci/install.ps1#L114

From the log output, it looks like Windoze is not liking a path starting with double quote, based on the error:

```
Cannot find drive. A drive with the name '"C' does not exist.
```

Note the double quote before the `C`. That being said, I don't know how to fix it, or what changed so that now it breaks 🙈

PR #277, opened 2023-06-21
Add Exchange version

Gsantomaggio:
Implement https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/153


Zerpet (commented):
Looks good so far. We have to store somewhere the response we get from `ExchangeVersions` command, so that we know what frame version we should use for `Deliver` and `Publish`. I think we should exchange versions shortly after opening successfully a connection to RabbitMQ.

Gsantomaggio:
@Zerpet yes, This PR is only the first step. Then I will implement the publish ( see #276 ) and the `deliver v2`

PR #303, opened 2023-09-19
Add docs to timestamp offset; add additional constructors

ricsiLT:
I guess at minimum I'd like docs merged - the additional constructors are optional but would help out.
Essentially, I consider existing constructor to be a bit of a foot-gun - I only realized that giving normal timestamp didn't work properly because I've read a bug report here :D


Zerpet (commented):
Thank you for the PR! It looks reasonable to me. What do you think @Gsantomaggio ?

Gsantomaggio:
Yes. Looks good to me! thanks @ricsiLT

Gsantomaggio (approved):

Gsantomaggio:
@ricsiLT do you think it is hard to create a test for that?

ricsiLT:
You mean, for passing in `timestamp` that's too small? I guess I could add it... like, to understand:

```
		var dto = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
		Console.WriteLine(dto); // Returns 1695207579
		
		var dto2 = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
		Console.WriteLine(dto2); // Returns 1695207579115
```

Do you want a test for the generic constructor of `OffsetTypeTimestamp(long timestamp)`, where I'd div the number by 10^10 and see if there's a remainder, and throw if user provides a timestamp that doesn't? Or you just want me to write a test where I pass a specific timestamp and see that it's indeed the correct numeric value?

Gsantomaggio:
@ricsiLT I meant a test that uses this filter, but don't worry, I will add it. I have to do some test refactoring.

PR #305, opened 2023-09-19
Force localhost connections when client connects to localhost

### LINKED ISSUE #296: Unable to create producer when connected to localhost and rabbit uses a hostname
### Describe the bug

From this conversation:  https://github.com/spring-projects/spring-amqp/issues/2522

When RabbitMQ is running in a environment where rabbit node name is not resolvable by the clients, and the client connects locally via the loopback, the client library fails to create a producer and retries infinitely.

This kind of setups are common for local development environments, where rabbit may resolve its node name as `foobar`, and the client connects via `localhost`. One example is running rabbit locally, and another, running rabbit in a container.

### Reproduction steps

1. Start RabbitMQ in a container, with a specific hostname
    ```
    docker run --hostname foobar --rm --name bunny -p 5672:5672 -p 15672:15672 -p 5552:5552 rabbitmq:3-management
    docker exec bunny rabbitmq-plugins enable rabbitmq_stream_management
     ```
2. Create a `StreamSystem` that connects to `localhost` and default port `5552`
3. Create a stream
4. Create a producer using `Producer.Create()`

Observe that the client does not create the producer connection, and it retries infinitely.

### Expected behavior

The client library connects via localhost, or gives up connecting.

### Additional context

Related https://github.com/rabbitmq/rabbitmq-server/issues/9424

The issue can be workarounded using an address resolver, or changing the advertised host/port to always be `localhost` and the default stream port. However, there's a fair point made in the Spring Boot discussion: https://github.com/spring-projects/spring-amqp/issues/2522#issuecomment-1717717363. In summary, if there's no load balancer involved (address resolved was created for LB setups), and there's no rabbit cluster, then the `localhost` and default loopback user configuration should work.

This may arguably be an improvement, rather than a bug.

Zerpet:
Fixes #296

It is common to have a local instance of RabbitMQ running locally,
either via a local rabbit (e.g. installed from source), or inside a
container (e.g. Docker or local Kubernetes). In both cases, our client
may not work without providing an address resolver, or without changing
the advertised host/port in rabbit via an env variable. Specifically,
our clients won't work if the client can't resolve the hostname of the
local machine.

For exampe, a laptop may have a hostname mylaptop.some-company.com.
Rabbit won't be impacted as long as it works in a single-node
configuration. However, the client smart features to locate the stream
leaders and replicas will get "confused" because it can't resolve the
hostname.

This commit changes the smart layer, so that it won't try to locate the
stream replicas if the client is connected to localhost. Instead, it
will force the producer/consumer connections to be localhost, and it
will assume that stream replicas/leader are in localhost. This is true
for single-node rabbits in local development environments. The smart
feature to locate leader/replicas remains unchanged for non-local
connections.

### DIFF
--- a/Tests/FilterTest.cs
+++ b/Tests/FilterTest.cs
@@ -18,10 +18,14 @@ public class FilterTest
 {
     // When the Filter is set also Values must be set and PostFilter must be set
     // Values must be a list of string and must contain at least one element
-    [Fact]
+    [SkippableFact]
     public async void ValidateFilter()
     {
         SystemUtils.InitStreamSystemWithRandomStream(out var system, out var stream);
+        if (!AvailableFeaturesSingleton.Instance.PublishFilter)
+        {
+            throw new SkipException("broker does not support filter");
+        }
 
         await Assert.ThrowsAsync<ArgumentException>(() => Consumer.Create(
             new ConsumerConfig(system, stream) { Filter = new ConsumerFilter() }
@@ -54,10 +58,14 @@ public class FilterTest
     // We send 100 messages with two different states (Alabama and New York)
     // By using the filter we should be able to consume only the messages from Alabama 
     // and the server has to send only one chunk with all the messages from Alabama
-    [Fact]
+    [SkippableFact]
     public async void FilterShouldReturnOnlyOneChunk()
     {
         SystemUtils.InitStreamSystemWithRandomStream(out var system, out var stream);
+        if (!AvailableFeaturesSingleton.Instance.PublishFilter)
+        {
+            throw new SkipException("broker does not support filter");
+        }
 
         var producer = await Producer.Create(
             new ProducerConfig(system, stream)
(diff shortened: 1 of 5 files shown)


Zerpet:
Requires some manual QA before merging.

@Gsantomaggio @lukebakken could we introduce this in a patch release? Arguably it doesn't add new functionality, but a slight change in behaviour.

Tests/FilterTest.cs
```diff
@@ -171,10 +179,15 @@ async Task SendTo(string state)
     // For the producer side we have the ConfirmationHandler the messages with errors 
     // will be reported as not confirmed and the user can handle them.
     // for the consumer the messages will be skipped and logged with the standard logger
-    [Fact]
+    [SkippableFact]
```
MikhailTushev, line 182: use `[Fact(Skip="reason")]`

Gsantomaggio (approved):
Looks good to me, should we do https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/305#discussion_r1330448734 ?

Tests/FilterTest.cs
Zerpet: I disagree. We only want to skip this test if the broker is < 3.13. We want to run always run if the broker is >= 3.13

Zerpet:
> Looks good to me, should we do [#305 (comment)](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/305#discussion_r1330448734) ?

I don't think we should mark the test to be skipped. In my commit, the test is marked "maybe skip", and it will always run, at least to the point where the `SkipException` is thrown.

Allow me to do manual QA for a couple hours before we consider merging.

Tests/FilterTest.cs
Gsantomaggio: ops, yes @Zerpet you are right

PR #308, opened 2023-09-21
Improve test speed

Gsantomaggio:
Speed up a bit the tests

- reduce the collect management time, it reduces the time to wait for the HTTP calls
- remove some wait since there is already the wait_until function
- remove the load definition and use the AMQP client to create the super-stream configuration


Tests/ReliableTests.cs
Zerpet, line 1: I'd rather remove commented code, we can always revert the commit if we want to "recover" this code.

Zerpet (commented):
Looking good 👍

PR #312, opened 2023-09-22
Removed unused class MyMethodAttribute from Producer.cs

TroelsL:
Found an unused internal class in Producer.cs. Unless there is a future use for this, it is not needed. But the naming seems to indicate that this was test code.


Gsantomaggio (approved):
Thank you

PR #313, opened 2023-09-23
Fix the Unsubscribe timeout problem

Gsantomaggio:
- The  Unsubscribe[ function sync](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/313/files#diff-b08ff9a3d581af9df4f7e209452830c93f2367debb487e937e158870332fbd1eR576) (like the others).  The consumers' list is updated in the right way.
The connection is correctly closed since there are no pending consumers.

- Fix the handle delivery if the consumer is removed, but there are still chunks on the wire.
Add debug [logs for this situation.](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/313/files#diff-71215640c5968601ad0a124946241692beb3c931cc450bb43bb7061e8f84c47eR468-R473) 

- Add Degug Asserts to validate the buffer size, which can help understand some parse chunk errors.
That can be temporary at some point; we can remove them.
Ref https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/310

- Add the [cancellation token](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/313/files#diff-c8897c227c93d047004a719f033ac3b57ea66a332af74f736684dd72c4320738R32) to the connection class
it helps when the [reader is blocked](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/313/files#diff-c8897c227c93d047004a719f033ac3b57ea66a332af74f736684dd72c4320738R142) and the consumer is closed.

- Given this [use case](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/310) so short-life consumers, the initial credits should be low to avoid keeping the read handler busy.

- Minor change is [replace Memory<T> to ReadOnlyMemory<T>](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/313/files#diff-1fb2e6a6fb7446b99c45579f5c6f5bb181cc32cd72ad83382379d2c69d6e23c0R54)  it is a bit faster and safe 

- Uniform the log debug and error messages  with [consumer info](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/313/files#diff-b08ff9a3d581af9df4f7e209452830c93f2367debb487e937e158870332fbd1eR125) it adds all the consumer info that can be helpful in case of debugging and errors

### DIFF
--- a/.ci/install.ps1
+++ b/.ci/install.ps1
@@ -52,7 +52,7 @@ $env:ERLANG_HOME = $erlang_home
 
 Write-Host "[INFO] Setting RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS..."
 $env:RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS = '-rabbitmq_stream advertised_host localhost'
-[Environment]::SetEnvironmentVariable('RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS', '-rabbitmq_stream advertised_host localhost', 'Machine')
+[Environment]::SetEnvironmentVariable('RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS', '-rabbitmq_stream advertised_host localhost -rabbit collect_statistics_interval 4', 'Machine')
 
 Write-Host '[INFO] Downloading RabbitMQ...'
 
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -45,7 +45,7 @@ jobs:
       rabbitmq:
         image: rabbitmq:3.13.0-beta.6-management
         env:
-          RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS: -rabbitmq_stream advertised_host localhost
+          RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS: -rabbitmq_stream advertised_host localhost -rabbit collect_statistics_interval 4
         ports:
             - 5552:5552
             - 5672:5672
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,6 @@ test: build
 rabbitmq-server:
 	docker run -it --rm --name rabbitmq-stream-docker \
 		-p 5552:5552 -p 5672:5672 -p 15672:15672 \
-		-e RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="-rabbitmq_stream advertised_host localhost" \
 		--pull always \
 		pivotalrabbitmq/rabbitmq-stream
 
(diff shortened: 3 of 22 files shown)


Gsantomaggio:
@jonnepmyra do you have a chance to test? 
It adds more detailed logs to understand the chunk parse error

jonnepmyra:
> @jonnepmyra do you have a chance to test? It adds more detailed logs to understand the chunk parse error

Thanks! I'm on it!

Zerpet (approved):

jonnepmyra:
I've been thoroughly testing this pull request over the past few days, and it appears to be a clear improvement.

Before this fix, we encountered the following errors during high-throughput scenarios with short-lived consumers. However, after applying this fix, these issues do not seem to occur anymore:

**Error while processing chunk: XYZ**
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at RabbitMQ.Stream.Client.RawConsumer.<>c__DisplayClass14_0.<<ParseChunk>g__DispatchMessage|1>d.MoveNext() in C:\git\rabbitmq-streamclient\rabbitmq-stream-dotnet-client\RabbitMQ.Stream.Client\RawConsumer.cs:line 201
```

**An error occurred while calling Dispose**
```
System.ArgumentNullException: Value cannot be null. (Parameter 'array')
   at System.Buffers.TlsOverPerCoreLockedStacksArrayPool`1.Return(T[] array, Boolean clearArray)
   at System.IO.Pipelines.BufferSegment.ResetMemory()
   at System.IO.Pipelines.BufferSegment.Reset()
   at System.IO.Pipelines.StreamPipeReader.CompleteAndGetNeedsDispose()
   at System.IO.Pipelines.StreamPipeReader.Complete(Exception exception)
   at RabbitMQ.Stream.Client.Connection.Dispose() in C:\git\rabbitmq-streamclient\rabbitmq-stream-dotnet-client\RabbitMQ.Stream.Client\Connection.cs:line 223
   at RabbitMQ.Stream.Client.Client.Close(String reason) in C:\git\rabbitmq-streamclient\rabbitmq-stream-dotnet-client\RabbitMQ.Stream.Client\Client.cs:line 583
```

While we are still experiencing some Close() timeouts with short-lived stream consumers, it's worth noting that this may be related to a non-default InitialCredits value (specifically, 100 instead of the default 2) mentioned by @Gsantomaggio. At present, we can consider this as a non-issue.

Great work and thanks for helping out!

Gsantomaggio:
Thank you @jonnepmyra for the contribution

Issue #315, opened 2023-09-25
Passing incorrect value to StreamSystem during creation.


TroelsL:
### Describe the bug

I may be incorrect, but it would appear the wrong client parameters are being passed on creation of StreamSystem. At least I had some difficulties getting it to work while setting up a test and connecting via IP.

Please see this diff:

https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/compare/main...TroelsL:rabbitmq-stream-dotnet-client:fix-endpoint-on-stream-system

### Reproduction steps

1. Setup a poorly configured RabbitMQ server in host mode networked docker on a remote server, bound to the default interface settings of 0.0.0.0:5552
2. Connect to Stream via a config like this:
```
var config = new StreamSystemConfig
{
    Password = "password",
    UserName = "user",
    Endpoints = new[] { new DnsEndPoint("hostname", 5552) }
};

var system = await StreamSystem.Create(config)!.ConfigureAwait(false);

if (!await system.StreamExists(StreamName).ConfigureAwait(false))
{
    await system.CreateStream(new StreamSpec(StreamName)).ConfigureAwait(false);
}

var producerConfig = new ProducerConfig(system, StreamName)
{

};
var producer = await Producer.Create(producerConfig, loggerFactory.CreateLogger<Producer>()).ConfigureAwait(false);
```
3. Observe that it fails to connect as it defaults to the advertised IP (127.0.0.1) endpoint despite having successfully tried and  connected to a specific endpoint.

### Expected behavior

I would expect the system to connect to the working endpoint over the advertised.

Gsantomaggio:
Does it work if you use `localhost` ?

TroelsL:
I tried reproducing my original issue, but I am currently unable. I also looked through the code, and I simply cannot find anywhere that the EndPoint stored on the StreamSystem is being used. As this is the case, I am closing this issue and removing the MR.

I now believe I was having unrelated issues, and the default EndPoint caused me to jump to invalid conclusions as I was debugging the code. It appears as though it was connecting to 127.0.0.1, probably due to this line in Client.cs:

```
public EndPoint Endpoint { get; set; } = new IPEndPoint(IPAddress.Loopback, 5552);
```

PR #319, opened 2023-09-28
Fixing Endpoint passed to StreamSystem

TroelsL:
This fix ensures that the clientParams passed to the StreamSystem matches the one set on the Client (and the one the connection has just been established to).

Fixes an issue where the server is bound to multiple IPs, and it advertises a local IP address over a public one.

Resolves: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/315

### DIFF
--- a/RabbitMQ.Stream.Client/StreamSystem.cs
+++ b/RabbitMQ.Stream.Client/StreamSystem.cs
@@ -66,12 +66,13 @@ namespace RabbitMQ.Stream.Client
             {
                 try
                 {
-                    var client = await Client.Create(clientParams with { Endpoint = endPoint }, logger)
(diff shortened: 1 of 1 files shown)


Gsantomaggio:
Thank you @TroelsL we are looking on it

Gsantomaggio:
It looks good to me, but I'd like to hear @Zerpet, who recently worked on https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/305, and I'd like to understand if there is any side effect. 


The endpoint is not passed since it is filled when there is the query metadata.

TroelsL:
Closing this PR due to a false positive. The change here will have no effect on the current implementation as the value passed set on ClientParameters.EndPoint in the context of StreamSystem is never used.

Issue #316, opened 2023-09-26
Implement client event bus  to expose the internal events 


Gsantomaggio:
### Is your feature request related to a problem? Please describe.

The client does a lot of automatic operation, and the user is not aware of what is happening.
We need to give more feedback to the user. 

### Describe the solution you'd like

The idea is to Implement an event bus to expose the internal events.
The client makes some decisions based on what we think is the best, for example in the Producer and Consumer classes, the auto-reconnect is more or less automatic, and the user is not aware of what is happening, see https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/discussions/314.

The client decides what to do in case of an error for example the CRC fail on some error parse.

We could expose events and the user can react to these events, ex stop the consumer af 3 reconnections or stuff like that.

cc @ricardSiliuk @ricsiLT @jonnepmyra @TroelsL

jonnepmyra:
Thank you for taking the time to consider our diverse needs and use cases for the library 😊. I believe it's a good idea to allow customization of the client to cater to specific use case requirements.

For instance, in many of ___our___ use cases, message loss is simply not acceptable, and a failed CRC check should immediately halt the consumer, rather than proceeding with the next chunk as it currently does. However, this approach may not be suitable for all scenarios.

In the event of a 'FailedCrcCheckEvent,' here's what we'd probably want to do:

1. Store the failed CRC event persistently, such as by incrementing the CRC failure count in Redis (e.g., IncrementCrcFailureInRedis(streamname, reference, (offset?), DateTime.UtcNow)).
 
2. Disconnect the consumer.
 
3. If the count of failed CRC checks for this stream/consumer/offset is less than 5 within the last hour, attempt to reconnect a new consumer at the last stored offset (before the CRC failed chunk). This assumes that the CRC failure was a transient issue, and hopefully, the chunk will pass the CRC check this time. If retries exceed 5, we will raise an alarm for our operations team.

Just want to make sure that the 'eventbus' waits for the event to be handled by our "custom eventhandler implementation" before proceeding to the next chunk. It's crucial that the consumer stops when the event is raised, preventing it from processing the next chunk (which leads to messageloss).

Gsantomaggio:
Closed in favour of https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/336

PR #320, opened 2023-10-04
Event bus [Proposal] 

Gsantomaggio:
ref: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/316

The idea of this implementation is to have the possibility to subscribe to a partial event.

so like:
```csharp
  StreamEventsBusSingleton.Instance.Publish<ProducerConnectEvent>(new ProducerConnectEvent(this, _config));
```

and somewhere to subscribe to the event based on the type, like:

```csharp
 StreamEventsBusSingleton.Instance.Subscribe<ProducerConnectEvent>(p =>
 {
   Console.WriteLine($"ProducerConnectEvent , Stream {p.Config.Stream}, event type {p.EventType}, event severity 
   {p.EventSeverity}" );
});
```

In this way we can implement different events with detailed information and (why not ) with the class instance that raised the event, for example:

```csharp
    public class ProducerConnectEvent : IStreamEvent
    {
        public ProducerConnectEvent(RawProducer producer, RawProducerConfig config)
        {
            Producer = producer;
            Config = config;
        }

        public RawProducerConfig Config { get; }

        public EventTypes EventType => EventTypes.Connection;
        public EventSeverity EventSeverity => EventSeverity.Info;
        public RawProducer Producer { get; }
    }
```


jonnepmyra:
Looks good.

We will need to execute async code when the events are raised.

Do you think the we could have async subscriptions aswell? So the registered action would be a `Func<IStreamEvent, Task> `and that the Publish part of the lib would `await` our registered callback/subscriber?

Gsantomaggio:
>have async subscriptions aswell? 

Yes

TroelsL:
I think this proposal covers our operational needs. But I am not certain which events the "Reconnecting" state would entail. From the WIP code, it looks like there are two event types: Connection and Disconnection.

I assume that, if modelled as a state machine, a Producer would start off in `Disconnected`. It would likely only have a single transition from there to `Connected`. But would there only be these two states, or would `Reconnecting` be it's own, and if so: would there be a "Reconnection" event published?

Gsantomaggio:
> it looks like there are two event types: Connection and Disconnection.

@TroelsL this was just a proposal. Of course, we will add other types like:
 - Reconnection  ( with all the details ) 
 - Metadata update 
 - CRC error
 - Producer Error
 - Consumer-Parse error 

etc...

Gsantomaggio:
It is still a work in progress. 
The bus can handle the events from the `RawProducer` and the `Producer` classes and only connection/disconnection/reconnection.

So given:
```
rabbitmq-streams add_super_stream invoices --partitions 3
```

And:
```csharp
 var streamSystem = await StreamSystem.Create(new StreamSystemConfig());
        await streamSystem.CreateStream(new StreamSpec("my-stream"));
        IEventBus bus = new StreamEventsBus();
        bus.Subscribe<RawProducerConnected>(async connected =>
        {
            myLogger.LogInformation(
                "The raw producer {ClientProvidedName} is connected to the stream {Stream}",
                connected.Parameters.ClientProvidedName, connected.Instance.Info.Stream);

            await Task.CompletedTask;
        });

        bus.Subscribe<RawProducerDisconnected>(async disconnected =>
        {
            myLogger.LogInformation(
                "The raw producer {ClientProvidedName} is disconnected from the stream {Stream}",
                disconnected.Parameters.ClientProvidedName, disconnected.Instance.Info.Stream);
            await Task.CompletedTask;
        });

        bus.Subscribe<ProducerReconnected>(async reconnected =>
        {
            var value = (reconnected.IsReconnection) ? "is in reconnection.." : "ended the reconnection";
            myLogger.LogInformation("The producer {ClientProvidedName} {Value} to the stream {Stream}", reconnected.Instance.Info.ClientProvidedName, value, reconnected.Instance.Info.Stream);
            await Task.CompletedTask;
        });

        var superStreamProducer = await Producer.Create(new ProducerConfig(streamSystem, SystemUtils.InvoicesExchange)
        {
            SuperStreamConfig = new SuperStreamConfig()
            {
                Routing = message => message.Properties.MessageId.ToString(),
                RoutingStrategyType = RoutingStrategyType.Hash
            },
            ClientProvidedName = "my_super_producer",
            Events = bus,
        });

        var standardProducer = await Producer.Create(new ProducerConfig(streamSystem, "my-stream")
        {
            ClientProvidedName = "my_producer",
            Events = bus,
        });


        for (ulong i = 0; i < 2000; i++)
        {
            var message = new Message(Encoding.Default.GetBytes("hello"))
            {
                Properties = new Properties() {MessageId = $"hello{i}"}
            };
            await superStreamProducer.Send(message);
            Thread.Sleep(1 * 1000);
        }
    }
```

The connection part for the super stream is:
```
[10:23:16] info: example.MyEventsTest[0] The raw producer my_super_producer#invoices-0 is connected to the stream invoices-0
[10:23:16] info: example.MyEventsTest[0] The raw producer my_super_producer#invoices-1 is connected to the stream invoices-1
[10:23:16] info: example.MyEventsTest[0] The raw producer my_super_producer#invoices-2 is connected to the stream invoices-2
```

for the standard stream is:
```
[10:23:16] info: example.MyEventsTest[0] The raw producer my_producer is connected to the stream my-stream
```

I case I kill the standard producer connection the events will be:
```
[10:23:45] info: example.MyEventsTest[0] The raw producer my_producer is disconnected from the stream my-stream
[10:23:45] info: example.MyEventsTest[0] The producer my_producer is in reconnection.. to the stream my-stream
[10:23:45] info: example.MyEventsTest[0] The raw producer my_producer is connected to the stream my-stream
[10:23:45] info: example.MyEventsTest[0] The producer my_producer ended the reconnection to the stream my-stream
```

The behaviour for the super stream is a bit different: the connection is re-created as soon a message is sent to the partition stream.


The basic idea is to define your bus with the subscriptions you want to listen to and handle it with different information. 

The bus is an interface `public interface IEventBus`.So the user can write its own event logger in case the default one is not enough.

Gsantomaggio:
closed in favour of https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/336

PR #322, opened 2023-10-10
Add Info class to the Producer/Consumer classes

Gsantomaggio:
The call returns the Reference and Stream.
So that you know, the `Reference` is deprecated for the producer class and will be removed at some point. So the `info.reference` for the producer will also be removed.

We can add other info about the consumer and producer and the consumer in the future.

ref: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/discussions/321 

To get the info:

```csharp
 var consumer = await Consumer.Create(new ConsumerConfig(system, "MyStream") { Reference = "consumer", });

entityInfo = consumer.Info;
Assert.Equal("MyStream", entityInfo.Stream);
Assert.Equal("consumer", entityInfo.Reference);
``` 

and also in the handler:

```csharp
MessageHandler = async (sourceStream, consumer, ctx, message) =>
{
                        
    Console.WriteLine($"Consumer info Reference: {consumer.Info.Reference} ");
```

### DIFF
--- a/RabbitMQ.Stream.Client/AbstractEntity.cs
+++ b/RabbitMQ.Stream.Client/AbstractEntity.cs
@@ -21,5 +21,7 @@ namespace RabbitMQ.Stream.Client
         }
 
         protected Client _client;
+
+        public Info Info { get; internal set; }
     }
 }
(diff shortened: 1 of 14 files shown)


Gsantomaggio:
FYI: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/discussions/321#discussioncomment-7250144

Zerpet (approved):

PR #329, opened 2023-11-21
Fix connections to server on different locales

Noonlord:
ToUpper is not enough, we need to use InvariantCulture to be sure that the string is converted to uppercase in a way that is consistent with the server.

### DIFF
--- a/RabbitMQ.Stream.Client/Client.cs
+++ b/RabbitMQ.Stream.Client/Client.cs
@@ -222,7 +222,7 @@ namespace RabbitMQ.Stream.Client
                     .ConfigureAwait(false);
             logger?.LogDebug("Sasl mechanism: {Mechanisms}", saslHandshakeResponse.Mechanisms);
 
-            var isValid = saslHandshakeResponse.Mechanisms.Contains(parameters.AuthMechanism.ToString().ToUpper(),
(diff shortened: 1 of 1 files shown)


Noonlord:
@Zerpet I would like to have a review if possible.

Gsantomaggio:
Thank you @Noonlord. 
May I ask for an example where the current implementation fails?

Zerpet (commented):
I echo Gabriele's thought. Where does the current implementation fail?

From what I read in [Microsoft documentation](https://learn.microsoft.com/en-us/dotnet/api/system.string.toupperinvariant?view=net-7.0), the upper invariant seems important for consistent sorting. However, we are not sorting the authentication mechanism array, we just iterate it.

Noonlord:
@Gsantomaggio although I can not provide you a configuration file at the moment, I can show something else to reproduce the bug locally.

This is the current culture:
resim (2)

This is the auth mechanisms response from server:
resim (1)

And this is the client parameters: 
resim

Issue is from client parameters authentication method enumeration, Plain, converts to string as "Plain" and using TR-tr culture ToUpper() method transforms this into "PLAİN" those two strings get compared and it results to false, because of this client thinks that server does not support "Plain" authentication method. 

Thanks for quick responses!

Gsantomaggio (approved):

Zerpet:
Thank you for the detailed explanation and the contribution!

Issue #340, opened 2023-12-21
Updating SuperStream Connection Access Token


Gsantomaggio:
### Discussed in https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/discussions/334

<div type='discussions-op-text'>

<sup>Originally posted by **lngvivek** December 13, 2023</sup>
We are using SuperStream dotnet client version 1.7.2
For Authentication with Rabbit, we are using IdenitityServer Acces Token. It works perfectly on initial connection. However once the access toke expires, we were looking for something UpdateSecret (as used to be in Queue based sdk) on Connection object so that our connection is updated with lastest Access token. Could not find any such option for updating token to RabbitMQ.Stream.Client.StreamSystem. Could anyone pls suggest on how to refresh the token?</div>

Gsantomaggio:
@simone-fariselli will follow this issue.

simone-fariselli:
Hi @Gsantomaggio,
I created the following method in the Client class:
```c#
internal async Task UpdateSecret()
{   
    var saslHandshakeResponse =
        await Request<SaslHandshakeRequest, SaslHandshakeResponse>(corr => new SaslHandshakeRequest(corr))
            .ConfigureAwait(false);
    _logger?.LogDebug("Sasl mechanism: {Mechanisms}", saslHandshakeResponse.Mechanisms);

    var saslData = Encoding.UTF8.GetBytes($"\0{Parameters.UserName}\0{Parameters.Password}");
    var authResponse =
        await Request<SaslAuthenticateRequest, SaslAuthenticateResponse>(corr =>
                new SaslAuthenticateRequest(corr, Parameters.AuthMechanism.ToString().ToUpperInvariant(),
                    saslData))
            .ConfigureAwait(false);
    ClientExceptions.MaybeThrowException(authResponse.ResponseCode, Parameters.UserName);
}
```

which should handle the change of secret (i.e of the access_token).

I get an applicative timeout and the following error log from the server:
```
unknown command {request,101,
                    {sasl_authenticate,<<"PLAIN">>,
                        <<0,103,117,101,115,116,0,103,117,101,115,116>>}}, sending close command.
Ignored unknown message emit_stats in state close_sent
Closing connection because of timeout in state 'close_sent' likely due to lack of client action.
unknown command {request,101,sasl_handshake}, sending close command.
Ignored unknown message emit_stats in state close_sent
[...]

PR #342, opened 2024-01-03
Update secret functionality

simone-fariselli:
add update secret functionality.
add test for UpdateSecret function.

Solves #340  with the introduction of the UpdateSecret method in StreamSystem. 
Usage example:
 ```csharp
await streamSystem.UpdateSecret("new_secret");
 ```


Gsantomaggio:
Thank you @simone-fariselli

Gsantomaggio (approved):

Issue #336, opened 2023-12-15
Improve reconnection


Gsantomaggio:
### Is your feature request related to a problem? Please describe.

The `Producer` and `Consumer` classes provide a way to auto-reconnect, but there is some edge case not correctly handled like:
- https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/335
- https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/discussions/333

The auto-reconnection is hidden from the user. It is hard to understand the status without logs. 

### Describe the solution you'd like

What to do:

- [x] Provide more feedback with events ( ref https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/316)
   via https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/349
- [x] Improve the reconnection when the stream is in a "limbo" status like `StreamNotAvaiable` which is different from `StreamDoesNotExist` done via https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/343
- [x] Stop to deliver messages as soon the consumer requests the closing.  It will avoid providing messages to the user that are not needed anymore (dove via https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/339)
- [x] Uniform the way to reconnect: The super stream has its own way to reconnect the client. We should use the same way ( ref  https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/discussions/333 )  dove via https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/344
- [x] Handle errors like:
```
[...]

PR #343, opened 2024-01-09
Improve Reconnection

Gsantomaggio:
* Part of https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/336

* Improve the disconnection message. It is possible to understand if it is a normal disconnection or an unexpected.
* Remove the Active Items from the connection pool and use the Publishers and Consumers client list directly to check the pool size
* Refactor the Factory Classes. Remove code duplication in case of metadata update and connection closed. See ReliableBase.OnEntityClosed
* Handle streamNotAvailable error. In this case the client will try to reconnect the entity
* Fix the events attach to the RawConsumer and RawProducer. The events are attached only if the ResponseCode is OK
* Detach the events once the entity receives the disconnection or metadata update. In that case, the entity is closed
* Introduce ReliableEntityStatus like a state machine to understand the status of Producer and Consumer classes
* Add ResourceAvailableReconnectStrategy interface to Handle the retry in case testing in a stream exists. See ReliableBase CheckIfStreamIsAvailable
* Change the  MetadataHandler to Func<MetaDataUpdate, Task> to be like the other methods
* Producer and Consumer classes fail fast during the first initialization. The user is aware of what is happening. The reconnect part occurs only after the first boot.

How to test
----
- Setup a cluster. I use:
```
git clone git@github.com:rabbitmq/rabbitmq-stream-go-client.git .
make rabbitmq-ha-proxy 
```

```
/etc/hosts
127.0.0.1 node0
127.0.0.1 node1
127.0.0.1 node2
```

- execute a script with multiple consumers and producers:
I use https://gist.github.com/Gsantomaggio/f6a091076e498d88ecfe827c2b165dad

- restart the docker nodes and see if the client will survive to the restart

### DIFF
--- a/RabbitMQ.Stream.Client/AbstractEntity.cs
+++ b/RabbitMQ.Stream.Client/AbstractEntity.cs
@@ -12,6 +12,7 @@ namespace RabbitMQ.Stream.Client
     public abstract record EntityCommonConfig
     {
         internal ConnectionsPool Pool { get; set; }
+        public Func<MetaDataUpdate, Task> MetadataHandler { get; set; }
     }
 
     internal enum EntityStatus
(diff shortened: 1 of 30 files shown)


lukebakken (dismissed):
Thanks for the tour @Gsantomaggio

Gsantomaggio:
@lukebakken Thanks for the help! [Lock removed](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/343/commits/10a3f5cb76793a895b3f3995357d5ddc028ef885)

Issue #351, opened 2024-02-05
Consumer connecting to leader node


francescopessina87:
### Describe the bug

As stated in the blog post "Connecting to Streams", a good practice for producers and consumers should be the following:
- producers should always connect to leader nodes
- consumers should always connect to follower nodes

(see https://blog.rabbitmq.com/posts/2021/07/connecting-to-streams/#well-behaved-clients)

Looking at RawConsumer.cs, it connects to the cluster using the `LookupRandomConnection` method (here: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/main/RabbitMQ.Stream.Client/RawConsumer.cs#L205)

As stated in the comment of the method (here: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/c3d8e555be3154f873aeb4d7a3229e97004cce28/RabbitMQ.Stream.Client/RoutingClient.cs#L178) that method "Gets a random connection. The consumer can connect to a replica or leader.", which is contradictory to the "well behaved clients" recommendation.

Furthermore, trying that with a cluster behind a load balancer, my consumers always connect to the stream leader node.

Could it be fixed to follow the recommended best practices?

### Reproduction steps

Connect a consumer to a cluster and check to which node the client is connected to. This should never be a stream leader node.

### Expected behavior

When a consumer connects to a stream in a cluster should connect to one of the "followers" node of a stream.

Gsantomaggio:
Grazie @francescopessina87 

Here the working in progress https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/352

PR #352, opened 2024-02-05
Consumer connected only to the followers

Gsantomaggio:
Fixes https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/351


Gsantomaggio:
@francescopessina87 do you have an opportunity to test it?

RabbitMQ.Stream.Client/RoutingClient.cs
```diff
@@ -180,8 +186,14 @@ await LookupConnection(clientParameters, metaDataInfo.Leader, MaxAttempts(metaDa
         public static async Task<IClient> LookupRandomConnection(ClientParameters clientParameters,
```
francescopessina87, line 186: @Gsantomaggio could it be a good idea to rename this method? We are no longer looking up for any "random" connection, but for a random "replica" connection (if any) or a "leader" connection... is that significant for the callers of this method?

francescopessina87, line 186: @Gsantomaggio I would also correct the comment specifying the new behaviour

Gsantomaggio: done via https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/352/commits/cfb6b241d4007a65d9ebf84d717f6afddfdd5b76

Gsantomaggio: done via https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/352/commits/cfb6b241d4007a65d9ebf84d717f6afddfdd5b76

francescopessina87 (approved):

Gsantomaggio:
Thank you @francescopessina87 ! Let's merge it!

Issue #363, opened 2024-03-06
On RabbitMQ restart, Updated Secret is not picked during connection recovery, resulting in broken publishers and subscribers


vivek-singh2:
### Describe the bug

We are using SuperStream and using RabbitMQ version 3.13.0, Erlang 26.2.2 and stream-dotnet-client 1.8.1
Authentication to RabbitMQ is being done using IDS. Update secret keeps on working until RabbitMQ server gets restarted. In case of restart (after original token expires) publisher/subscribers do not reconnect. Please see Reproduction steps for exact details.

Please use below repo to test this-

https://github.com/lngvivek/UpdateSecretIssueTest/tree/main/Test

### Reproduction steps

Token Life time is 15 min.

1. At 14:30, Connections were created (Subscribers and Publishers)
2. At 14:35 restarted Rabbit then Subscribers and Publishers recovered successfully.
3. At 14:37 (Secret was refreshed successfully. Refreshed it few mins before expiry)
4. AT 14:47 (after 17 mins of original connection, restarted Rabbit Again. This time consumer/pub do not reconnect and remain disconnected forever)

### Expected behavior

Subscribers and Publisher connections should be recoverd.

Gsantomaggio:
It could fail until you will exedatepdate secrets again. 
You can check the  [status change event here](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/main/docs/ReliableClient/BestPracticesClient.cs#L207) for producer and consumer.

Play with these events can help to request the secret. 
Note: Don't put blocking code on the events

PR #364, opened 2024-03-06
Updated Secret try to reconnect the client 

Gsantomaggio:
fixes https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/363

The new password is set before trying to reconnect


vivek-singh2:
Thanks @Gsantomaggio 
Tested latest commit by breaking connection with RabbitServer (Didn't restart Rabbit Server) Pub/Sub connections were recovered.

PR #367, opened 2024-03-14
Add missing ReadUuid.

bangjiehan:
Read UUID as 16-length byte array.


bangjiehan:
There might be some endianness problem about .Net GUID struct so return raw byte array just for now.

Gsantomaggio:
Thanks for the PR @bangjiehan .
May I ask to ask to add a test for it? If you have a way to generate a message with UUID you can do like I did here: 
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/main/Tests/Amqp10Tests.cs#L300

>struct so return raw byte array just for now.

I Agree with that

Gsantomaggio (approved):

Gsantomaggio:
Thank you @bangjiehan !

PR #368, opened 2024-03-18
Read VersionString from AssemblyInformationalVersionAttribute (#365)

doerig:
Read VersionString from AssemblyInformationalVersionAttribute instead of FileVersionInfo


lukebakken (approved):
Thank you @doerig! We appreciate contributions from the community!

Issue #371, opened 2024-04-11
Receiving AmqpParseException: Read Any: Invalid type: 208 after upgrading to RabbitMQ 3.13.1


ngbrown:
### Describe the bug

I upgraded a RabbitMQ cluster to 3.13.1 and my client using `RabbitMQ.Stream.Client` version `1.8.2` started throwing exceptions:

> Error while parsing message "Consumer id 0 for stream: stream-overview, identifier: , reference: table-consumer-stream-overview, OffsetSpec RabbitMQ.Stream.Client.OffsetTypeOffset Client ProvidedName dotnet-stream-consumer-stream-overview-00d3, No SuperStream, IsSingleActiveConsumer: False, Token IsCancellationRequested: False ".  The message will be skipped. Please report this issue to the RabbitMQ team on GitHub "https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues"

with the following stack trace:

```
RabbitMQ.Stream.Client.AMQP.AmqpParseException: Read Any: Invalid type: 208
   at RabbitMQ.Stream.Client.AMQP.AmqpWireFormatting.ReadAny(SequenceReader`1& reader, Object& value) in /_/RabbitMQ.Stream.Client/AMQP/AmqpWireFormattingRead.cs:line 65
   at RabbitMQ.Stream.Client.AMQP.Map`1.Parse[T](SequenceReader`1& reader, Int32& byteRead) in /_/RabbitMQ.Stream.Client/AMQP/Map.cs:line 31
   at RabbitMQ.Stream.Client.RawConsumer.<>c__DisplayClass14_0.<ParseChunk>g__MessageFromSequence|0(ReadOnlySequence`1& unCompressedData, Int32& compressOffset) in /_/RabbitMQ.Stream.Client/RawConsumer.cs:line 266
```

### Reproduction steps

Connect to a stream on RabbitMQ 3.13.1, where the data is bound to an exchange that is either shoveled or federated with another cluster that is providing the data on it's own exchange (3.12.x in this case). Clients are providing data to the exchange on the source cluster.

### Expected behavior

Work correctly with new versions of RabbitMQ.

### Additional context

I did not enable any new feature flags after upgrading. All feature flags were enabled in 3.12.12. I did turn on the configuration `classic_queue.default_version = 2` as part of the upgrade. (Tried again without this configuration, doesn't make a difference).

Gsantomaggio:
Can you please elaborate more?
[...]

PR #372, opened 2024-04-11
Update ReadAny to skip over lists and maps

ngbrown:
Addresses #371 enough to at least get the client working again with RabbitMQ 3.13.

For now, this just returns `null` as the value for lists and maps. The project can decide how to correctly handle this, since the `x-received-from` message annotation has the following structure:

```json
{
  "x-received-from": [
    {
      "uri": "amqp://something",
      "exchange": "events",
      "redelivered": true,
      "cluster-name": "rabbit@something",
      "vhost": "./"
    }
  ]
}
```

The `x-shoveled` header has the following structure:

```json
{
  "x-shovelled": [
    {
      "shovelled-by": "rabbit@820c50d2107b",
      "shovel-type": "dynamic",
      "shovel-name": "events-from-...",
      "shovel-vhost": "./",
      "src-uri": "amqp://somewhere",
      "dest-uri": "amqp://",
      "src-exchange": "events",
      "src-exchange-key": "#",
      "dest-exchange": "events"
    }
  ]
}
```

But since the message annotation of AMQP 1.0 is completely flexible, this client shouldn't lock into just the current value types. Maybe return generic `List<object>` and `Dictionary<string, object>`?


Gsantomaggio:
According to the AMQP 1.0 specification, ' application properties` are only flat values; `annotations` can be flat and list.

https://www.rabbitmq.com/docs/conversions#amqpl-amqp

The current parser handles all the maps in the same way. We'd need to distingue when we are parsing:
 `application properties`  or `annotations` in the first case your fix is correct. In the second case, we should handle it ( like you are doing in the other branch ) 




thank you for the fix

Gsantomaggio (approved):
It is ok for the Application Properties

Gsantomaggio:
FYI for the moment : https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/releases/tag/v1.8.3 

Thank you for the fix

PR #380, opened 2024-05-22
remove the LastOffset()

Gsantomaggio:
The method name may be misleading. It does not indicate the last offset of the stream. It indicates the last uncommited chunk id. This information is not necessary. The user should use CommittedChunkId().


Gsantomaggio:
@lukebakken, I'm not sure why it happened. It changed the file as soon as I downloaded the repo.

Issue #388, opened 2024-07-09
Improve Connection Management for Short-Lived Consumers


jonnepmyra:
### Is your feature request related to a problem? Please describe.

This PR https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/328 significantly improved the time required to create consumers to a stream, provided the shared connection is not drained from active consumers. However, in scenarios where the creation pace of short-lived consumers is not constant, and there is a time gap between the creation of each consumer, we still experience a relatively long delay. I guess this happens because the first created consumer is closed before the second one is opened, and therefore the underlying connection used for consumer #1 will be closed since it does not have any active consumers.

### Example scenario: 
When handling incoming HTTP requests, we connect a consumer to a RabbitMQ stream to rebuild a state. While receiving a few hundred messages is swift, the time-consuming connection creation process results in sluggish HTTP response times.

### Describe the solution you'd like

The root cause of this delay is that the underlying connection is closed when there are no active consumers. To mitigate this, we propose maintaining the underlying connection open, even in the absence of active consumers. This approach mirrors the behavior of AMQP connections, which can exist without active channels or consumers

### Additional context

_Originally discussed in https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/328#issuecomment-1838567089_

PR #389, opened 2024-07-12
Add Connection Pool Close configuration

Gsantomaggio:
Closes: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/388

### DIFF
--- a/RabbitMQ.Stream.Client/ClientExceptions.cs
+++ b/RabbitMQ.Stream.Client/ClientExceptions.cs
@@ -38,7 +38,8 @@ namespace RabbitMQ.Stream.Client
                 return x.Any();
             }
 
-            return exception is (SocketException or TimeoutException or LeaderNotFoundException or InvalidOperationException or OperationCanceledException) ||
+            return exception is (SocketException or TimeoutException or LeaderNotFoundException
+                       or InvalidOperationException or OperationCanceledException) ||
                    IsStreamNotAvailable(exception);
         }
 
@@ -187,6 +188,14 @@ namespace RabbitMQ.Stream.Client
     public class TooManyConnectionsException : Exception
     {
         public TooManyConnectionsException(string s)
+            : base(s)
+        {
+        }
+    }
+
+    public class PendingConnectionsException : Exception
+    {
+        public PendingConnectionsException(string s)
             : base(s)
         {
         }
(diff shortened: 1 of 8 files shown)


Gsantomaggio:
@jonnepmyra Can you please test it?
```csharp
                ConnectionPoolConfig = new ConnectionPoolConfig()
                {
                    ConnectionCloseConfig = new ConnectionCloseConfig()
                    {
                        Policy = ConnectionClosePolicy.CloseWhenEmptyAndIdle,
                        IdleTime = TimeSpan.FromSeconds(60)
                    }
                }
```

jonnepmyra:
Wow, that was fast :) , thank you!

I'm afraid that I will not be as fast ;),  it may take a little longer than usual to test due to the holiday period, but I'll test as soon as possible!

Gsantomaggio:
Thanks a lot @lukebakken 

Here https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/1814584722be71c15adfe41cefdeb4827725287a/RabbitMQ.Stream.Client/ConnectionsPool.cs#L212

You raised a good point and an old "problem" of this client:
When you close the `StreamSystem,` the producers and consumers created by `StreamSystem` won't be closed. 

The behaviour is different from Java and Go, see for [example, Go Stream Client](https://github.com/rabbitmq/rabbitmq-stream-go-client/blob/baed3d514424e0924142231059167ad1c1598f5b/pkg/stream/environment.go#L264-L269 
) 
 

We should also add this behaviour to this client, but it will introduce a sort of breaking change. Can we think of a 1.9 version? 

The change has to deal with the auto-reconnect feature and close all the producers and consumers in the right way to avoid triggering the auto-reconnection

Gsantomaggio:
@lukebakken To avoid inconsistency, I added a check during the [pool connection close](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/389/files#diff-c0155414eac5347211d45756c09892d1488460eb4ba5ca6944260d95243867b9R431)  only in case the policy  is `CloseWhenEmptyAndIdle` 

That is not perfect, but it could be enough for the moment.

Version 2.0, we can think of adopting the same Golang and Java behaviour: 
```
Close all the Producers and Consumers when the System is closed. 
```
This is a breaking change that I want to avoid introducing in this version. 


With the policy `CloseWhenEmpty` ( default ) won't change anything btw. 

@jonnepmyra thoughts?

PR #390, opened 2024-07-25
Update RabbitMQ version and the test

Gsantomaggio:
Update the test conversion from amqp 091 to stream due of https://github.com/rabbitmq/rabbitmq-server/pull/11715


Gsantomaggio:
@lukebakken green again :)! I will merge!

PR #391, opened 2024-08-28
Fix Client memory leak caused by not detaching UnhandledException event handler

Jordan-Osborn:
The event handler registered in the client's constructor closes over the client, and is never detached. The AppDomain.CurrentDomain UnhandledException event invocation list lives for the length of the program and acts as a GC root for any created clients (due to the addition of this handler), this means clients will get leaked even if they are no longer reachable from elsewhere in the program.

In this PR I have just removed the registration of the handler, I think it's fairly unusual for a client library to need to register a handler to receive notification for all unhandled exceptions in the current app domain. But please let me know if you disagree.

(On a separate but related note, I have also noticed that the underlying socket in Connection is not deterministically cleaned up, for example if Client.Create throws, connection is not disposed of and we rely on the finalizer of Socket to clean up underlying resources (at the moment the finalizer never runs because of the reference chain Client -> Connection -> Socket and the event handler never being detached)

### DIFF
--- a/RabbitMQ.Stream.Client/Client.cs
+++ b/RabbitMQ.Stream.Client/Client.cs
@@ -59,7 +59,6 @@ namespace RabbitMQ.Stream.Client
         public delegate Task MetadataUpdateHandler(MetaDataUpdate update);
 
         public event MetadataUpdateHandler OnMetadataUpdate;
-        public Action<Exception> UnhandledExceptionHandler { get; set; } = _ => { };
         public TimeSpan Heartbeat { get; set; } = TimeSpan.FromMinutes(1);
 
         public string ClientProvidedName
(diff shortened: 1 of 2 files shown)


Gsantomaggio:
Thank you @Jordan-Osborn. Yes, you are right, the `  AppDomain.CurrentDomain.UnhandledException` should be removed from there. 

cc @lukebakken

Gsantomaggio (approved):

Gsantomaggio:
Thank you @Jordan-Osborn 

Can I ask to open another issue with:

> (On a separate but related note, I have also noticed that the underlying socket in Connection is not deterministically cleaned up, for example if Client.Create throws, connection is not disposed of and we rely on the finalizer of Socket to clean up underlying resources (at the moment the finalizer never runs because of the reference chain Client -> Connection -> Socket and the event handler never being detached)


If you can add some details, that would be useful.

Issue #393, opened 2024-10-09
No Connection Resilience when missing too many heartbeats


bastl98:
### Describe the bug

Currently, the stream connection is not resilient when too many heartbeats are missed and there is a timeout on the connection close. 
All publishes of this connection result in a client timeout error and all consumers of this connection stop working.

There is also no reconnection attempt, because this error is processed as a "normal" connection close. A manual reconnection attempt is also not possible, because all properties which could be used to check if there is a need for a reconnection attempt still indicate that the connection is available:
![image](https://github.com/user-attachments/assets/25c0e736-97f9-4b1e-aeed-f9a2bd561e5a)

### Reproduction steps

1.Start RMQ Cluster with 3 nodes (docker)
2.Create a stream system and a consumer with the client lib which connects to one of the nodes (lower heartbeat interval for testing)
![image](https://github.com/user-attachments/assets/8dc7bc33-fca3-4352-8f4e-9dca288487f7)
3.Pause the the node to which the consumer is connected
4.Wait for 4 Heartbeats
5. The following errors are produced by the client lib
![image](https://github.com/user-attachments/assets/d2ed0f64-b407-46cc-be89-4bf5c97df279)
6. The consumer stops working but the method .IsOpen() still returns true

### Expected behavior

The timeout error should not result in a "normal" connection close, this should lead to a reconnection attempt by the library itself.

### Additional context

Our RMQ Cluster has version 3.13.0 and our client lib has version 1.8.2

bastl98:
I have adapted the BestPracticesClient for reproduction purposes (Logging in Consumer Callback and lowered hearbeat timespan). I have attatched the BestPracticesClient and appsettings.json which i have used to reproduce the error.

Here´s the link to the repo: [https://github.com/bastl98/rmq-bug-source.git](url)

Steps:
1. Start the client and wait until the consumer starts consuming
2. Pause the rmq container the consumer is connected to and wait until the heartbeat threshhold is reached
3. The connection is closed "normal" because of a timeout in the connection close procedure
4. When restarting the conatiner, the consumer and the connection will not recover but the consumer is still open

Gsantomaggio:
@bastl98 Thank you for reporting the issue.

The library is working properly. 
It is precisely the scope for the heartbeat to close the client when the client does not receive the "alive" from the server. 

The problem here is the Consumer with the status `IsOpen == true` is even closed and should be set as closed.
The correct status should be `IsOpen == false`

[...]

PR #394, opened 2024-10-17
Handle too many heart missing with reconnection 

Gsantomaggio:
Close from normal to Unexpected.
Fixes: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/393

Changes:

Remove the update status form the the Dispose() function.
It is unnecessary since the disposal can be called only from the Client close() function
where the status is set.

Add a function to convert the connection close reason string to the Reliable.ChangeStatusReason Enum.
This mapping should be improved to version 2.0, where we can introduce breaking changes.
In version 1.x, we need the conversion function.


Gsantomaggio:
@bastl98 can you please try the PR? the client should reconnect producer and consumer

lukebakken (commented):
Great!

PR #402, opened 2025-02-25
Reduce memory usage

SoftStoneDevelop:
Usage:

```csharp
var producerConfig = new DeduplicatingProducerConfig(...)
{
    ConfirmationHandler = async confirmation =>
    {
        // call dispose on each confirmation.Messages or do nothing
    },
};

var producer = await DeduplicatingProducer.Create(producerConfig, ...);
// start producer...

// send new message
var memory = MemoryPool<byte>.Shared.Rent(10);
// fill memory...

await producer.Send(
    messageId,
    new Message(memory, 10));
```

### DIFF
--- a/RabbitMQ.Stream.Client.PerfTest/Program.fs
+++ b/RabbitMQ.Stream.Client.PerfTest/Program.fs
@@ -40,7 +40,7 @@ let main argv =
         let! producer = system.CreateRawProducer producerConfig
         //make producer available to metrics async
         prod <- producer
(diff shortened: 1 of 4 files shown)


Gsantomaggio:
Thank you for your PR. I will check it as soon as possible

Gsantomaggio:
@SoftStoneDevelop, may I ask for a report on the memory gain? We'd like to see some numbers.  We will change the `Message` class, which is critical for the library and we have to be sure. 
Thank you.

PR #406, opened 2025-03-23
Add partitions to the StatusInfo

Gsantomaggio:
The partition field is deprecated now. The event status can be raised for more partitions, for example, in case of super stream open or close.
Fix the case where the super stream status moves from init to open, and the partition is empty.

@jonnepmyra Can you try it?


jonnepmyra:
Thanks, i'm on it!

Gsantomaggio:
@lukebakken I did the test we discussed. I Replaced the DLL in release mode without recompiling the module, and everything works fine.  

So, I think we can stay on `1.8.x`.

lukebakken (approved):

jonnepmyra:
Thanks for this fix, it works great, all the statuschange scenarios with superstreams is covered now

PR #408, opened 2025-03-26
Expose Store offset to stream system 

Gsantomaggio:
Expose Store offset to stream system


Gsantomaggio:
@jonnepmyra can you format the code?

jonnepmyra:
Some background to this PR....

In a message handler for a `Consumer`, you have access to the consumer instance, which allows you to store offsets on the stream—this works well within the context of receiving messages.

However, there are use cases where offset storage is needed outside of the message handling flow. For example when the application is shutting down and we want to explicitly call `StoreOffset()`.

This PR addresses that need

bangjiehan:
That's what I need. I try to process messages in batch so I saved and updated a reference to RawConsumer(I'm not sure. It might be disposed later, right?) just for calling StroeOffset. Now I could get rid of it and just call StoreOffset by StreamSystem.

Issue #370, opened 2024-03-20
Avoid throwing OffsetNotFoundException in StreamSystem.QueryOffset(string, string)


bangjiehan:
### Is your feature request related to a problem? Please describe.

I think only when there indeed is something wrong does C# throw a exception. Or at least there is some way to check before, for example call File.Exists before calling File.Read~.
There seems no way to check whether Offset exist or not before calling QueryOffset.

### Describe the solution you'd like

Add one of the following overloads or modify current implementation of QueryOffset.
```
bool exist = StreamSystem.TryQueryOffset(string, string, out ulong offset)
ulong? offset = StreamSystem.TryQueryOffset(string, string)
```

Gsantomaggio:
Thank you @bangjiehan 
I think you are right. 

We can add  `bool exist = StreamSystem.TryQueryOffset(string, string, out ulong offset)`  (it is more idiomatic)

The important thing is to keep the current implementation. We could mark it as deprecated and remove it on `2.0`

PR #413, opened 2025-04-08
Add TryQueryOffset API

Gsantomaggio:
Closes: https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/issues/370

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -18,8 +18,8 @@
     <PackageVersion Include="coverlet.collector" Version="3.2.0" />
     <!-- docs/**/*.csproj -->
     <PackageVersion Include="K4os.Compression.LZ4.Streams" Version="1.2.16" />
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="7.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="8.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="8.0.0" />
   </ItemGroup>
   <ItemGroup Label="net8.0 specific" Condition="'$(TargetFramework)' == 'net8.0'">
     <!-- RabbitMQ.Stream.Client -->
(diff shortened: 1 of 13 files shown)


Gsantomaggio:
@bangjiehan Better late than never :) WDYT?

bangjiehan:
Thank you for trying to improve this API.

TryQueryOffset could be implemented without throwing any Exception at all. Maybe put actual implementation in TryQueryOffset(exception-free, of course) and reimplement QueryOffset by wrapping TryQueryOffset(if returned value is null, thrown exception to maintain old behavior).
Throwing exception is a somehow expensive operation in C#. It create exception object, capture calling stack, search exception handlers etc. So, in normal flow, we won't throw and capture an exception to process a value we could expect.

I found StreamStats.FirstOffset() and StreamStats.CommittedChunkId() have similar issue.

Gsantomaggio:
Thank you @bangjiehan 
I can't touch `QueryOffset` due to compatibility. 

So you would do:
```csharp
  try
            {
                var qOffset = await QueryOffset(reference, stream).ConfigureAwait(false);
                return qOffset;
            }
            catch (Exception)
            {
                return null;
            }
```

Isn't that misleading? I mean, any Exception? 

  
About QueryOffset and TryQueryOffset I have to leave both the API for at least a reasonable time

Gsantomaggio:
>I found StreamStats.FirstOffset() and StreamStats.CommittedChunkId() have similar issue.

I can't remove these API I can Add other API(s) with TryFirstOffset and TryCommittedChunkId

thank you for let me know

Gsantomaggio:
@bangjiehan ok now `TryQueryOffset` is excpetion Free. 

@lukebakken WDYT?

bangjiehan:
@Gsantomaggio
Sorry, my last comment was misleading and wrong. I did not notice there are other `Exception`s could be thrown by different `ResponseCode`. I was thinking only about `OffsetNotFoundException`.

My original intention is below:

```
public async Task<ulong> QueryOffset(string reference, string stream)
{
    ulong? offset = TryQueryOffset(reference, stream);
    return offset ?? throw new OffsetNotFoundException($"QueryOffset stream: {stream}, reference: {reference}"));
}

// Mostly copied from original QueryOffset.
public async Task<ulong?> TryQueryOffset(string reference, string stream)
{
    MaybeThrowQueryException(reference, stream);

    var response = await _client.QueryOffset(reference, stream).ConfigureAwait(false);

    // Offset do not exist so just return null. There is no need to throw an OffsetNotFoundException and capture.
    if (response.ResponseCode = ResponseCode.OffsetNotFound)
    {
        return null;
    }

    ClientExceptions.MaybeThrowException(response.ResponseCode, $"QueryOffset stream: {stream}, reference: {reference}");
    return response.Offset;
}
```

Gsantomaggio:
@bangjiehan makes more sense now! I will merge thank you

PR #414, opened 2025-04-15
Improve socket handling 

Gsantomaggio:
- Improve socket handling by using the native API `socket.Connected` instead of a local variable to check if the connection is still open or not 
- Handle the `close` workflow in a better way 
- Add 
```csharp
  reader = PipeReader.Create(stream, new StreamPipeReaderOptions(leaveOpen: true));
  writer = PipeWriter.Create(stream, new StreamPipeWriterOptions(leaveOpen: true));
``` 

- Add a `try catch` block to the message handler; in case it fails for any reason, it will be logged. 
- Remove one test. There was a bug, and it was not deterministic.  (Many other tests cover the same functions )

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -18,8 +18,8 @@
     <PackageVersion Include="coverlet.collector" Version="3.2.0" />
     <!-- docs/**/*.csproj -->
     <PackageVersion Include="K4os.Compression.LZ4.Streams" Version="1.2.16" />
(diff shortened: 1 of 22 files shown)


Gsantomaggio:
@jonnepmyra, do you have a chance to test it?

jonnepmyra:
> @jonnepmyra, do you have a chance to test it?

Yes, but won't be able to test until Thursday next week 😔

PR #417, opened 2025-05-27
lock update listener

Gsantomaggio:


jonnepmyra:
This seems to do the trick -Thanks for the update!

PR #433, opened 2025-09-05
Guard PipeReader.AdvanceTo on closed socket

moikot:
Problem: A race during shutdown could call reader.AdvanceTo(buffer.Start, buffer.End) after the TCP socket is closed, risking exceptions that propagate via reader.CompleteAsync(caught) and noisy logs.
Change: In Connection.ProcessIncomingFrames, only call reader.AdvanceTo(...) when socket.Connected is true.

### DIFF
--- a/RabbitMQ.Stream.Client/Connection.cs
+++ b/RabbitMQ.Stream.Client/Connection.cs
@@ -189,7 +189,10 @@ namespace RabbitMQ.Stream.Client
                         numFrames += 1;
                     }
 
-                    reader.AdvanceTo(buffer.Start, buffer.End);
+                    if (socket.Connected)
+                    {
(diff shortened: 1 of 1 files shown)


moikot:
I'm not sure how to test this though, but I see that every time I close a consumer an InvalidOperationExcetion is thrown from `reader.AdvanceTo`

Gsantomaggio:
the code looks good btw!

moikot:
@Gsantomaggio Thank you for reviewing the fix and pointing me to the README! I'll start the process immediately!

Gsantomaggio (approved):

Gsantomaggio:
Thanks @moikot

moikot:
Thank you, @Gsantomaggio, for the prompt review and for accepting my PR ☺️

PR #446, opened 2026-01-26
ConsumerFactory: use exception overload for logger

gboucher90:


Gsantomaggio:
close due of inactivity

